PR #13, opened 2015-10-07
(maint) Add ability to override the Gemfile on dependency modules

cyberious:
- Use pattern established by puppetlabs and specify our defaults
  in the config default rather than hard coding the Gemfile

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -12,3 +12,21 @@
     env: PUPPET_GEM_VERSION="~> 3.0" FUTURE_PARSER="yes"
   - rvm: 2.1.6
     env: PUPPET_GEM_VERSION="~> 4.0" STRICT_VARIABLES="yes"
+Gemfile:
+  required:
+    ':test':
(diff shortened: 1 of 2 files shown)


blkperl:
:-1:  This increases the code complexity

igalic:
it's probably the right path, but i'm not happy with the test failures…

should we exclude the gemfile?

cyberious:
@blkperl we are either going to manage the Gemfile or have a bunch of modules we don't manage it on resulting in different patterns. I would rather have the complexity here than to have snowflakes everywhere. This is more relevant when talking about mixing modules for Windows and Linux systems where the dependencies vary.

cyberious:
@igalic I will look into the failures

igalic:
@cyberious :+1: for adding complexity in :one: place.

:bomb: :snowflake:

igalic:
fixed by #37!
thanks @cyberious && @jyaworski!

PR #29, opened 2015-10-19
We now manage puppet-rundeck

jyaworski:
Puppet-rundeck is ready to be managed


igalic:
\o/ yay!
thank you so much for this @jyaworski!

PR #40, opened 2015-11-13
Use puppet-lint from github

alexjfisher:
There hasn't be a release for a *long* time, but there have been some
fixes since then.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -18,6 +18,8 @@ Gemfile:
       - gem: rake
       - gem: rspec-puppet
         git: https://github.com/rodjek/rspec-puppet.git
+      - gem: puppet-lint
+        git: https://github.com/rodjek/puppet-lint.git
       - gem: metadata-json-lint
       - gem: rspec-puppet-facts
       - gem: rspec


cyberious:
My only concern is that how do we determine breaking changes, I think it would be better if we can get @rodjek to ship a new release if one is needed with semvar

alexjfisher:
A new release would be better, but I'm not expecting one any time soon. :(
https://github.com/rodjek/puppet-lint/issues/401
https://github.com/rodjek/puppet-lint/issues/443

rspec-puppet is set to the latest git version even though there have been new releases of that.
https://github.com/puppet-community/modulesync_config/blob/master/config_defaults.yml#L20

For puppet-lint at least, how about fixing on the latest commit instead? https://github.com/rodjek/puppet-lint/commit/2546fed6be894bbcff15c3f48d4b6f6bc15d94d1

cyberious:
Commented on it as well, perhaps it is time for PuppetLabs to step up and take over it.

igalic:
@alexjfisher this needs a rebase now, although, i think it might already be merged in a different pr

juniorsysadmin:
:+1:

PR #42, opened 2015-12-06
Align configuration a bit closer to the Puppetlabs one

juniorsysadmin:
- Use a broader .gitignore
- Use config includes/extras/allow_failures and don't hardcode
  rvm/puppet versions in .travis.yml
- Use .rspec rather than rspec.opts
- Don't hardcode disabled lint options

In addition:
- Reduce the default Gem/Puppet set to recent versions
- Be slightly more lenient on lint checks

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -1,17 +1,14 @@
 ---
 .travis.yml:
-  script: "\"bundle exec rake validate lint spec SPEC_OPTS='--format documentation'\""
   includes:
-  - rvm: 1.8.7
-    env: PUPPET_GEM_VERSION="~> 3.0"
-  - rvm: 1.9.3
-    env: PUPPET_GEM_VERSION="~> 3.0"
-  - rvm: 2.1.5
-    env: PUPPET_GEM_VERSION="~> 3.0"
-  - rvm: 2.1.5
-    env: PUPPET_GEM_VERSION="~> 3.0" FUTURE_PARSER="yes"
-  - rvm: 2.1.6
-    env: PUPPET_GEM_VERSION="~> 4.0" STRICT_VARIABLES="yes"
+  - rvm: 2.1.7
+    env: PUPPET_GEM_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test
+  - rvm: 2.1.7
+    env: PUPPET_GEM_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=test
+  - rvm: 2.2.3
+    env: PUPPET_GEM_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=test
+  - rvm: 2.2.3
+    env: PUPPET_GEM_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=rubocop
 Gemfile:
   required:
     ':test':
@@ -39,4 +36,12 @@ Gemfile:
     ':system_tests':
       - gem: beaker
       - gem: beaker-rspec
+      - gem: beaker-puppet_install_helper
+Rakefile:
+  default_disabled_lint_checks:
+  - 'relative'
+  - 'disable_80chars'
+  - 'disable_class_inherits_from_params_class'
+  - 'disable_documentation'
+  - 'disable_single_quote_string_with_variables'
 ...
--- a/moduleroot/.gitignore
+++ b/moduleroot/.gitignore
@@ -1,6 +1,16 @@
-.*.sw?
-pkg
-spec/fixtures
-.rspec_system
-.vagrant
+pkg/
 Gemfile.lock
+vendor/
+spec/fixtures/
+.vagrant/
+.bundle/
+coverage/
+log/
+.idea/
+*.iml
+.*.sw
+<% if ! @configs['paths'].nil? -%>
+<% @configs['paths'].each do |path| -%>
+<%= path %>
+<% end -%>
+<% end -%>
--- a/moduleroot/.travis.yml
+++ b/moduleroot/.travis.yml
@@ -1,6 +1,4 @@
 ---
-notifications:
-  email: false
 sudo: false
 language: ruby
 cache: bundler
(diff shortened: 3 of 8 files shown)


moduleroot/Rakefile
```diff
@@ -7,16 +7,12 @@ require 'rubocop/rake_task'
...
-PuppetLint.configuration.send('disable_80chars')
 PuppetLint.configuration.log_format = '%{path}:%{linenumber}:%{check}:%{KIND}:%{message}'
 PuppetLint.configuration.fail_on_warnings = true
-
-# Forsake support for Puppet 2.6.2 for the benefit of cleaner code.
-# http://puppet-lint.com/checks/class_parameter_defaults/
-PuppetLint.configuration.send('disable_class_parameter_defaults')
-# http://puppet-lint.com/checks/class_inherits_from_params_class/
-PuppetLint.configuration.send('disable_class_inherits_from_params_class')
+<% checks = @configs['default_disabled_lint_checks'] + ( @configs['extra_disabled_lint_checks'] || [] ) -%>
```
liamjbennett: Is the a specific reason we are not using a `puppet-lint.rc` file for this?

moduleroot/.travis.yml
```diff
@@ -1,6 +1,4 @@
 ---
-notifications:
```
rnelson0: I don't think we want emails going out for ever Travis run, that would probably just lead to people bit-bucketing them anyway

rnelson0:
Where are the new default lint checks?

What's the functional net change here, and is there a reason to move closer to puppet labs vs puppet-module-skeleton?

moduleroot/Gemfile
```diff
@@ -1,4 +1,4 @@
-source ENV['GEM_SOURCE'] || 'https://rubygems.org'
+source ENV['GEM_SOURCE'] || "https://rubygems.org"
```
igalic: this change was made because of rubocop complaining…
which, frankly, i find rather silly. imo, there should be only one string type…

juniorsysadmin:
> Where are the new default lint checks?

They're missing, I left them off by accident!

juniorsysadmin:
> What's the functional net change here, and is there a reason to move closer to puppet labs vs puppet-module-skeleton?

I'm trying to add a bit more flexibility to syntax checks and the gem/Puppet sets (I believe at the moment these are hardcoded). The idea being that once something has been modulesync-d with defaults the barrier to passing these checks/tests is low.

moduleroot/Gemfile
juniorsysadmin: I wondered about this, I will revert it back to how it was.

moduleroot/Rakefile
juniorsysadmin: Very good point. I had heard (perhaps wrongly) that sometimes options were ignored in puppet-lint.rc. Can modulesync currently template a puppet-lint.rc file without any dramas?

config_defaults.yml
```diff
@@ -3,15 +3,17 @@
   script: "\"bundle exec rake validate lint spec SPEC_OPTS='--format documentation'\""
```
juniorsysadmin: This line is currently ignored and not used in travis.yml at all. Do we care about --format documentation? Should we kill this line and continue to hardcode ```script``` in travis.yml, or actually template this line in travis.yml?

moduleroot/.gitignore
```diff
@@ -1,6 +1,15 @@
-.*.sw?
```
rnelson0: `.*.sw?` should definitely be kept.

rnelson0:
Interpreting the runtime contents of .gitignore/.travis.yml/Gemfile/Rakefile appears a bit more difficult when done this way. Perhaps I am just more comfortable with the `puppet-module-skeleton` format because of familiarity, though.

I would also recommend the use of PUPPET_GEM_VERSION over PUPPET_VERSION; factor's variable is done that way and that is used in `puppet-module-skeleton` which would make it more consistent with other modules people may maintain or test (like me! definitely a selfish request!).

moduleroot/Rakefile
juniorsysadmin: @liamjbennett Correct me if I'm wrong, but ```puppet-lint.rc``` doesn't seem to allow ignore_paths to be put in there

moduleroot/Gemfile
juniorsysadmin: Reverted

moduleroot/.gitignore
juniorsysadmin: Fixed

juniorsysadmin:
> Where are the new default lint checks?

>> They're missing, I left them off by accident!

Should be fixed now

juniorsysadmin:
> I would also recommend the use of PUPPET_GEM_VERSION over PUPPET_VERSION;
I agree, but I think this should be left to another PR, since this one is already creeping in scope.

juniorsysadmin:
On the most recent commit I have culled Ruby versions < 2.x, and also have been a bit more lenient with regard to default lint checks.

igalic:
huh? why killing off ruby < 2.x?
i'm fairly certain we have a great number of modules that need support for it
(some of us probably even still support 1.8.7)

rnelson0:
We are past EOL for [1.8.7](https://www.ruby-lang.org/en/news/2014/07/01/eol-for-1-8-7-and-1-9-2/) and [1.9.3](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/), and coming up on a full year since 1.9.3 was EOL'ed. It seems like as good a time as any to cut the cord, but perhaps giving 1.9.3 another extension wouldn't be horrible.

rnelson0:
That said, there's a suspicious lack of Ruby 2.0.x tests in the matrix.

juniorsysadmin:
> huh? why killing off ruby < 2.x?
> i'm fairly certain we have a great number of modules that need support for it
> (some of us probably even still support 1.8.7)

That's right, but we should require less effort for modules which don't want to support Ruby < 2.0, (no .sync.yml needed). If a module needs 1.8.7/1.9, here's an example .sync.yml that goes in the root of the repo:
https://github.com/puppetlabs/puppetlabs-ntp/blob/master/.sync.yml

PR #48, opened 2015-12-22
fix (rvm) bring back 1.9.3 testing

igalic:
while we have decided some time ago to discontinue our testing of 1.8.7,
we have not made such a decision for 1.9.3. And i believe, despite the fact
that 1.9.3 is eol, too, we *cannot* make that move just yet.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -1,6 +1,8 @@
 ---
 .travis.yml:
   includes:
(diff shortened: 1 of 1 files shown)


igalic:
merging for now, will be reverted once the world moves on…
please also note the conspicuous lack of ruby 2.0

PR #51, opened 2016-01-04
pull puppet blacksmith from github until released

igalic:


igalic:
https://github.com/voxpupuli/puppet-blacksmith/pull/24 is in blacksmith master, but as of yet unreleased. we use in our travis_release task

daenney:
:+1:

PR #65, opened 2016-01-22
Removes deprecated argument

petems:
Fixes 
```
Error: The `Style/TrailingComma` cop no longer exists. Please use `Style/TrailingCommaInLiteral` and/or `Style/TrailingCommaInArguments` instead.
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,7 +11,7 @@ Metrics/LineLength:
   Max: 120
 
 # dealbreaker:
-Style/TrailingComma:
+Style/TrailingCommaInArguments:
   Enabled: false
 
 # we still support ruby 1.8


daenney:
When was that changed?

petems:
rubocop 0.36 IIRC

jyaworski:
Hello:

Modulesync has settled on 0.35.0 for now, so this is not needed.

daenney:
We should look into upgrading though but yes, currently this won't affect anything: https://github.com/voxpupuli/modulesync_config/blob/master/config_defaults.yml#L32-L33

jyaworski:
Right; unneeded at this time, but good. Does this work in 0.35.0?

jyaworski:
Merging.

PR #68, opened 2016-01-25
Add support for latest releases of CentOS, Debian and Ubuntu

dhoppe:
Please update the nodesets, because the ones from [Puppetlabs](https://github.com/puppetlabs/modulesync_configs/tree/master/moduleroot/spec/acceptance/nodesets) are obsolete. They also changed the ```box_url```, since the boxes are hosted [Atlas](https://vagrantcloud.com/puppetlabs/).


juniorsysadmin:
I believe you can drop the ```box_url``` unless it's not hosted on Atlas and recent Vagrant versions should operate fine.
Also:
Did you mean to drop Centos 5.x support?
Is there a Ubuntu 15.04 image we can use as well?

juniorsysadmin:
By the looks of it, it sadly doesn't look like we can use https://github.com/puppetlabs/beaker-hostgenerator to generate these yet.

dhoppe:
@juniorsysadmin I dropped the config for CentOS 5, because it is EOL next year and I assumed that everybody is using CentOS 6. 

There is no Vagrant box available for Ubuntu 15.04, because it is not an LTS release.

jyaworski:
@dhoppe while I'm a huge fan of dropping CentOS5, Puppet's model is to support everything until it's EOL (at least). Should we deviate from that?

dhoppe:
@juniorsysadmin @jyaworski Alright, you got me. I will add support for CentOS 5.11 and remove the ```box_url```.

jyaworski:
LGTM. Merging.

PR #70, opened 2016-01-28
Add 3.x future parser to test matrix

jyaworski:
@igalic @daenney @nibalizer


juniorsysadmin:
Why? That's what testing with Puppet 4 is for. Future parser also lags behind Puppet 4 parsing fixes.

bastelfreak:
:+1: from my side, I know a few environments that run on puppet3 + future parser

daenney:
@juniorsysadmin Not entirely true. The upgrade path from 3 to 4 is
* Upgrade to 3.8
* Turn on future parser
* Upgrade to 4

Being able to check 3.8 + future parser is very useful.

PR #74, opened 2016-02-04
Add support for code coverage

dhoppe:
Please merge. This will produce coverage reports, to make sure that we have for almost everything a RSpec check.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,6 +3,7 @@ AllCops:
     - ./**/*.rb
   Exclude:
     - moduleroot/spec/spec_helper.rb
+    - moduleroot/spec/classes/coverage_spec.rb
     - vendor/**/*
     - modules/**/*
 
--- a/moduleroot/spec/classes/coverage_spec.rb
+++ b/moduleroot/spec/classes/coverage_spec.rb
@@ -1,0 +1,3 @@
+require 'rspec-puppet'
+
+at_exit { RSpec::Puppet::Coverage.report! }


jyaworski:
Tests fail?

dhoppe:
I am pretty sure this is caused by PR #65 and Rubocop 0.35.0, because the error message is complaining about an unrecognized cop ```Style/TrailingCommaInArguments```.

Since we are using a different parameter at [modulesync](https://github.com/voxpupuli/modulesync_config/blob/master/.rubocop.yml#L14) and [moduleroot](https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.rubocop.yml#L39), I recommend to use ```Style/TrailingComma``` or upgrading to Rubocop 0.37.0.

Both solutions solved the issue during local tests.

juniorsysadmin:
> I recommend to use Style/TrailingComma or upgrading to Rubocop 0.37.0.

Can we do this in a separate PR then, before merging this?

juniorsysadmin:
On an unrelated note, I don't think it's a good idea to create branches on the voxpupuli namespace rather than your own one, unless you intend that others make changes directly to the branch or you want others to finish the PR off.

dhoppe:
@juniorsysadmin Alright. Just take a look at #75.

dhoppe:
@juniorsysadmin Please merge this PR too.

igalic:
oooh! i got excited for a second thinking you added https://github.com/tushartushar/Puppeteer

jyaworski:
I think I'll look at implementing Puppeteer....

PR #75, opened 2016-02-08
Pin rubocop to version 0.37.0

dhoppe:
This should solve the error message ```unrecognized cop Style/TrailingCommaInArguments``` of issue #74.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ source "https://rubygems.org"
 
 group :test do
   gem 'rake'
-  gem 'rubocop', '0.35.0'
+  gem 'rubocop', '0.37.0'
 end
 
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
It might be worth doing a modulesync of some of the modules in managed_modules.yml to see if this version introduces some gotchas

dhoppe:
@juniorsysadmin In this case I have to pin Ruby to version 2.1.7, because 2.1.8 broke every bulid at Travis CI in the last few days.

juniorsysadmin:
Ah right, I'm okay with merging this then.

dhoppe:
@juniorsysadmin Then you should push the magic button.

PR #78, opened 2016-02-18
Adding more Puppet versions in travis tests

aerostitch:
Hi,

There have been significant changes in several versions of Puppet 3.
It's a good practice to test the different versions.

For example, when I build a dependent package with Travis on puppet 3.5 to 3.8 I sometimes get facts that disappeared from the standard.

If you could just tell me if you'd be ok to add those environments to the travis builds, it would be awesome.

Thanks! :)
Joseph

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -10,6 +10,10 @@
   - rvm: 2.1.8
     env: PUPPET_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test FUTURE_PARSER=yes
   - rvm: 2.1.8
+    env: PUPPET_VERSION="~> 3.5" STRICT_VARIABLES="yes" CHECK=test FUTURE_PARSER=yes
+  - rvm: 2.1.8
+    env: PUPPET_VERSION="~> 3.8" STRICT_VARIABLES="yes" CHECK=test FUTURE_PARSER=yes
+  - rvm: 2.1.8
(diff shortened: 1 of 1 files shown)


config_defaults.yml
```diff
@@ -10,6 +10,10 @@
   - rvm: 2.1.8
     env: PUPPET_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test FUTURE_PARSER=yes
   - rvm: 2.1.8
+    env: PUPPET_VERSION="~> 3.5" STRICT_VARIABLES="yes" CHECK=test FUTURE_PARSER=yes
```
jyaworski: Future parser shouldn't be used before 3.8.

Furthermore, it didn't exist in 3.5.

jyaworski:
I want to say that we've said that 3.8 is the only 3.x we're supporting going forward, but I'd like other opinions on that.

bastelfreak:
Also, if you want to test 3.5, the syntax isn't correct. ~> 3.X will always go to the latest 3.X release, 3.8.6 currently. So this commit would add redundancy to the matrix.

I've the same opinion as @jyaworski, I would prefer to only test against the latest 3.X release. Puppetlabs provides repos for these packages and it should be easy for everybody to upgrade from any 3.X to the latest 3.X.

aerostitch:
You're totally right, I don't know why I totally forgot that ~>3.0 means the latest 3.X available. This was totally stupid. To push this. Sorry.
Ok if you're policy is only to support the latest 3.X then it's settled, I'll close this! :)

PR #79, opened 2016-02-27
Use rubocop 0.37.2

jyaworski:


juniorsysadmin:
I would prefer that Rubocop get bumped in modulesync less frequently than it is at the moment. There are often significant changes between minor version bumps, and sometimes a number of bugs.

jyaworski:
@juniorsysadmin my thinking was that we should always be at the latest minor version on whatever major version we're using. That's why I did this without much fanfare, reading the changelogs.

juniorsysadmin:
@jyaworski Oh I didn't realise it was a patch and not 'minor' release (in semver terms) from what we had currently.

PR #86, opened 2016-03-01
Change to get user to leave trailing comma

petems:
Raise an error if users don't leave trailing commas. Nicer for the coder who comes after you.
This might be a bit controversial, it's still kinda argued in the Ruby community: https://github.com/bbatsov/ruby-style-guide/issues/273


nibalizer:
No, I don't think we need to do this.

petems:
Ok, can we change `TrailingCommaInLiteral` to disabled for consistency then?

jyaworski:
I'm on the fence about this, leaning against. I see the arguments for (mainly, easier reordering), but the arguments against are also valid (visual clutter).

PR #89, opened 2016-03-02
Improved conditional release

rnelson0:
There are two builds on rvm 1.9.3. One will pass tests, upload to the forge, and result in a success. The other will pass tests, try to upload to the forge and find the release already exists, and result in a failure. To prevent this, add a condition to match only one of the two 1.9.3 builds.


rnelson0:
See https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A for documentation. I don't know how to test this ahead of time, though.

PR #96, opened 2016-03-11
Use the latest Puppet 3 version

rnelson0:


jyaworski:
Can you also do this to rubocop? Picking up point releases is safe.

rnelson0:
An alternative format would be `gem "puppet", ENV['PUPPET_GEM_VERSION'] || '~> 3.0'` - I think that's clearer to read, but I stuck with the original format.

moduleroot/Gemfile
```diff
@@ -54,7 +54,7 @@ else
 gem 'facter', :require => false, :groups => [:test]
 end
 
-ENV['PUPPET_VERSION'].nil? ? puppetversion = '3.8.6' : puppetversion = ENV['PUPPET_VERSION'].to_s
+ENV['PUPPET_VERSION'].nil? ? puppetversion = '>~ 3.0' : puppetversion = ENV['PUPPET_VERSION'].to_s
```
jyaworski: Wrong direction?

config_defaults.yml
```diff
@@ -36,7 +36,7 @@ Gemfile:
       - gem: voxpupuli-release
         git: https://github.com/voxpupuli/voxpupuli-release-gem.git
       - gem: rubocop
-        version: '0.38.0'
+        version: '~> 0.38'
```
igalic: really hope this doesn't come back to bite us, since rubocop is still 0.x, and hence, semver-free-for-all

rnelson0: This should result in 0.38.z, NOT 0.38-0.99999999, so unless rubocop gets even WORSE about point releases, we should be relatively safe.

igalic: here's hoping ;)

PR #97, opened 2016-03-13
Change character width to 140

dhoppe:
Please merge. 

As you can see at [PR #419](https://github.com/rodjek/puppet-lint/pull/419) the character width has been changed to 140 and our [current configuration](https://github.com/voxpupuli/modulesync_config/blob/master/config_defaults.yml#L61) has no effect.


igalic:
~_~ oh that's where that's coming from

PR #108, opened 2016-04-02
Bump module default Ruby 2.1.x version to 2.1.9

juniorsysadmin:


juniorsysadmin:
rvm doesn't seem to know about 2.1.9 yet though...

jyaworski:
Do we have the ability to use the ~> syntax here, somehow?

PR #111, opened 2016-04-12
Start testing against jruby

jyaworski:
Start testing against jruby per http://juanitofatas.com/2015/02/25/travis-ruby-template/

Also, use the latest ruby versions instead of pinning to a specific one.

@daenney @igalic @rnelson0

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -1,24 +1,20 @@
 ---
 .travis.yml:
   includes:
-  - rvm: 1.9.3
+  - rvm: jruby-19mode
+    env: PUPPET_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test JRUBY_OPTS="-J -Xmx1024M --debug"
+  - rvm: jruby-19mode
+    env: PUPPET_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=test JRUBY_OPTS="-J -Xmx1024M --debug"
+  - rvm: 2.1
     env: PUPPET_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test
-  - rvm: 2.1.8
-    env: PUPPET_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test
-  - rvm: 1.9.3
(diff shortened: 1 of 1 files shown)


rnelson0:
What benefit does jruby provide us, and does it outweigh the increased time for tests to complete with the additional test matrix entries? It's a 30% increase in tests and probably at least a 30% increase in time, unless jruby is faster.

electrical:
There is no real advantage of testing against jruby. It need to start the JVM so by default its slower.

jyaworski:
Puopetserver uses jruby. We should test against it to be complete. I myself have run into problems with jruby 1.9.3 vs ruby 1.9.3.

moduleroot/.travis.yml
```diff
@@ -21,7 +21,10 @@ addons:
     <%- end -%>
   <%- end -%>
 <%- end -%>
-before_install: rm Gemfile.lock || true
+before_install:
+  - rm Gemfile.lock || true
+  - gem update --system || true
```
daenney: Why?

moduleroot/.travis.yml
```diff
@@ -21,7 +21,10 @@ addons:
...
+  - gem --version || true
```
daenney: Why?

rnelson0:
Why is 2.2 now tested against 4.0 but not 3.0, but 2.3 is tested against only 3.0? Perhaps you should re-sort the matrix by either puppet version, or ruby version, rather than the jumble it is now, to make it clearer what is and is not being tested.

moduleroot/.travis.yml
jyaworski: I could likely do without this; it was a best-practices thing I read.

moduleroot/.travis.yml
jyaworski: I could likely do without this; it was a best-practices thing I read.

jyaworski:
@rnelson0 is there a canonical 'we should test against this' list?

rnelson0:
Let's discuss the matrix questions in #113 and restrict comments here to the code review.

PR #117, opened 2016-04-15
Add puppetlabs-strings initial support

jyaworski:
@rnelson0 @bastelfreak

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -28,6 +28,8 @@ Gemfile:
         git: https://github.com/rodjek/rspec-puppet.git
       - gem: puppet-lint
         git: https://github.com/rodjek/puppet-lint.git
+      - gem: puppet-strings
+        git: https://github.com/puppetlabs/puppetlabs-strings.git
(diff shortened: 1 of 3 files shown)


juniorsysadmin:
I *strongly* dislike PRs that sneak in more than one thing. Do a PR for the Ruby bumps (if they're available yet from rvm) then do this one.

moduleroot/.pmtignore
```diff
@@ -0,0 +1 @@
+vendor/
```
juniorsysadmin: Puppet docs say:
> If you have both a .pmtignore and a .gitignore file, the Puppet module tool will read the .pmtignore file over the .gitignore

My preference: stick with .gitignore

jyaworski:
@juniorsysadmin done.

moduleroot/.pmtignore
rnelson0: FYI there are issues with PMT and it does not actually use .gitignore in many situations. Just something to keep in mind if we see module sizes exploding.

PR #125, opened 2016-04-23
pin listen on jruby and ruby21

bastelfreak:
one of our needed gems has listen as a dependency. they upgraded listen
from 3.0.6 to 3.1.1. It needed at least ruby19, now it's ruby22. We try
to pin it to the last working version on certain platforms. this is
based on:
https://github.com/theforeman/puppet-puppet/blob/8c0ccfc8268d90f8ae19d92a4d395733a845db84/Gemfile
https://github.com/theforeman/foreman-installer-modulesync/blob/b4ce08e08933b6dcf88978a50304e056cd1878aa/config_defaults.yml

for reference:
https://travis-ci.org/voxpupuli/puppet-collectd/jobs/125300240#L206


bastelfreak:
@jyaworski could you take a look? I'n not completely sure about the syntax of the jruby version

config_defaults.yml
```diff
@@ -46,6 +46,12 @@ Gemfile:
       - gem: puppet-lint-classes_and_types_beginning_with_digits-check
       - gem: puppet-lint-unquoted_string-check
       - gem: puppet-lint-variable_contains_upcase
+      - gem: listen
+        version: '3.0.6'
+        options:
+          platforms:
+            - 'ruby_21'
+            - 'jruby_19'
```
jyaworski: This looks like jruby-19mode.

https://github.com/voxpupuli/puppet-collectd/blob/master/.travis.yml#L12-L13

rnelson0:
That is horrible for a minor release. Has anyone reported this to listen as breaking server yet?

bastelfreak:
I opened https://github.com/guard/listen/issues/374

bastelfreak:
for reference: https://github.com/mojombo/semver/issues/148

bastelfreak:
so, after digging very deeply in very dirty places, the awesome @jyaworski and I came to two working solutions:

We can get the needed behavior with a little if statement:
```ruby
install_if -> { Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.2.0') } do
  gem 'listen', '~> 3.0.0', :require => false
end
```

@jyaworski [discovered that](https://github.com/voxpupuli/puppet-collectd/pull/476#issuecomment-214056855), I [tested it](https://travis-ci.org/voxpupuli/puppet-collectd/builds/125581691) . However this is hard to automate if we need to do this for multiple gems in the future, the only solution that came up my mind is to hardcode these three lines in our Gemfile provided by modulesync_config. This sucks because not every module needs the listen gem.

---

the way more awesome method:
```ruby
gem 'listen', '3.0.6', :require => false, :platforms => [:ruby_21, :jruby]
```
adding a platforms attribute to the gem line works fine, I [tested this](https://travis-ci.org/voxpupuli/puppet-collectd/builds/125582840). The modulesync_config update in this PR should add the above code, however this is currently untested. It would allow us to specify and ruby/gem version combination that we want. 

---

Both of these solutions require a recent rubygems/bundler version. Getting them will be implemented in https://github.com/voxpupuli/modulesync_config/pull/127. I [tested that](https://travis-ci.org/voxpupuli/puppet-collectd/builds/125581691) already. @jyaworski @rnelson0 can I get some opinions here? If you prefer the templated solution (the second one), like I do, I would like to rebase/squash this PR, test if it generates the correct Gemfile and merge it. Thank you all!

jyaworski:
Squash please.

rnelson0:
Merging this with just the option to pin to `:platform`, rather than actually pinning anything. We will hold off on that until a user actually requires it.

PR #126, opened 2016-04-24
hardcode needed listen version on <ruby2.2

bastelfreak:
this is a long story, please have a look at
https://github.com/voxpupuli/modulesync_config/pull/125


jyaworski:
We'll need to compare with Gem::Version.new.

PR #127, opened 2016-04-25
Update the system gems before doing anything else

jyaworski:
@bastelfreak

### DIFF
--- a/moduleroot/.travis.yml
+++ b/moduleroot/.travis.yml
@@ -21,7 +21,13 @@ addons:
     <%- end -%>
   <%- end -%>
 <%- end -%>
-before_install: rm Gemfile.lock || true
+before_install:
+  - bundle -v
+  - rm Gemfile.lock || true
+  - gem update --system
+  - gem update bundler
+  - gem --version
+  - bundle -v
 script:
   - 'bundle exec rake $CHECK'
 matrix:


rnelson0:
I dare say, we should solve these issues in A module before pushing it to modulesync configs. We should also link to an issue or provide a description of the what and why behind the changes.

jyaworski:
@rnelson0 while toying with #126 @bastelfreak and myself discovered that the version of bundle Travis uses is ancient (1.7.6). Almost all of us develop on the latest version (1.11.2). The same could be said with rubygems. That's what this PR is fixing.

bastelfreak:
As a side note, only old ruby versions ship bundle 1.7.6, ruby2.2 and 2.3 don't have this issue.

moduleroot/.travis.yml
```diff
@@ -21,7 +21,12 @@ addons:
     <%- end -%>
   <%- end -%>
 <%- end -%>
-before_install: rm Gemfile.lock || true
+before_install:
+  - bundle -v
+  - rm Gemfile.lock || true
+  - gem update --system
```
bastelfreak: do we really want to update everything or just bundler? This could bring some sideeffects.

bastelfreak: this is recommended by travis. I'm fine with this.

bastelfreak:
For reference, I tested this successful multiple times in https://github.com/voxpupuli/puppet-collectd/pull/476

rnelson0:
What is broken that is being fixed? What is the performance impact this has?

Being a little overly pedantic here but would like to ensure that we are tracking why we do things here, and that anyone using modulesync on their own can review the voxpupuli configs and understand if they need the same changes in their own configs.

bastelfreak:
sure sure :)

@jyaworski could you update the commit message with some infos? We've to install certain version of gems on certain ruby platforms. This is detailed explained in https://github.com/voxpupuli/modulesync_config/pull/125 . There are several ways to declare the the needed ruby + gem version in a Gemfile. the nice ones require a bundler version that isn't present in every default env from travis.

bastelfreak:
We also need to update bundler. Can you add the line like [here](https://github.com/voxpupuli/puppet-collectd/pull/476/commits/d5b1b18df0cf1deece4b8785f6222d6fca9d2394)?

jyaworski:
@bastelfreak done.

bastelfreak:
I'm fine with this. @rnelson0 are you too?

rnelson0:
What is the performance impact this has? Is this necessary for all modules or just when you need 'listen' to be 3.0.6, and if the latter, when is that?

jyaworski:
@rnelson0 this is brought on by the 'listen' thing, but not confined to it. This brings only performance benefits and brings our testing tooling in-line with our dev tooling. We all test on the latest rubygems/bundle, so Travis should run that.

bastelfreak:
@jyaworski same opinion here. I only checked a single run, gem update took 4.7s, bundler update 4.1. Gem install time went down from 14s to 12s.

jyaworski:
@bastelfreak @rnelson0 I believe this can be merged given the above metric.

PR #132, opened 2016-04-26
Allow rake target 'test' to be configured

traylenator:
Currently the rake target *test*  is hardcoded to
metadata_lint, lint, syntax and spec.

This list is now configured via a
*default_enabled_rake_targets* and an optional
*extra_disabled_rake_targets* that removes rake
targets from from the *test* target.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -62,4 +62,10 @@ Rakefile:
   - 'disable_class_inherits_from_params_class'
   - 'disable_documentation'
   - 'disable_single_quote_string_with_variables'
+  default_enabled_rake_targets:
+  - 'metadata_lint'
+  - 'lint'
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi,
what is the usecase for this? You've got any specific tests that you want to add on a certain module?

traylenator:
I'm attempting to use some of the same modulesync_configs internally all be it with gitlab CI.
Some modules not in a maintained state have broken rspec tests and I'd rather avoid
that problem for now. I've no intention of using *extra_disabled_rake_targets* within *voxpupuli* but
I would like to benefit from updates to the Rakefile.

jyaworski:
This is fine; it may open up new ideas in the future.

PR #146, opened 2016-05-06
Remove redundant gem, reorder

alexharv074:
Puppetlabs_spec_helper pulls in rake (also pulls in rspec-puppet and
puppet-lint, but these seem to be pointing to the tip of the github
projects for a reason).

Also places gems in a more natural order.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -21,24 +21,13 @@
 Gemfile:
   required:
     ':test':
-      - gem: rake
+      - gem: puppetlabs_spec_helper
       - gem: rspec-puppet
         git: https://github.com/rodjek/rspec-puppet.git
+      - gem: rspec-puppet-facts
+      - gem: rspec-puppet-utils
       - gem: puppet-lint
         git: https://github.com/rodjek/puppet-lint.git
-      - gem: puppet-strings
-        git: https://github.com/puppetlabs/puppetlabs-strings.git
-      - gem: metadata-json-lint
-      - gem: rspec-puppet-facts
-      - gem: rspec
-      - gem: puppet-blacksmith
-        git: https://github.com/voxpupuli/puppet-blacksmith.git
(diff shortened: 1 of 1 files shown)


config_defaults.yml
```diff
@@ -21,31 +21,29 @@
...
       - gem: puppet-lint-classes_and_types_beginning_with_digits-check
       - gem: puppet-lint-unquoted_string-check
       - gem: puppet-lint-variable_contains_upcase
+      - gem: metadata-json-lint
+      - gem: puppet-blacksmith
+        git: https://github.com/voxpupuli/puppet-blacksmith.git
+      - gem: voxpupuli-release
+        git: https://github.com/voxpupuli/voxpupuli-release-gem.git
+      - gem: puppet-strings
+        git: https://github.com/puppetlabs/puppetlabs-strings.git
```
rnelson0: Is there a reason we're pulling 1puppet-lint`, `puppet-blacksmith`, `voxpupuli-release`, and `puppet-strings` from git instead of the published gems? If we're doing iterative development more recent than the gems, seems like that would be appropriate in a module, rather than in our global config.

alexharv074: I also wondered about that.

jyaworski: @rnelson0 with puppet-lint specifically, it's because there hasn't been a release in.... a while.

puppet-strings is in flux. @danzilio may know more. So we use the git version.

alexharv074: I'm no doubt peaching to choir, but a better solution would be for the maintainers to release Gems.  I'll see if I can push them along to do this.

igalic: voxpupuli-release-gem hasn't been released to rubygems (yet)
we should change that.

alexharv074:
If others want to upvote https://github.com/rodjek/puppet-lint/issues/443

PR #148, opened 2016-05-07
Remove some redundant requires from Rakefile

alexharv074:
All of these are in puppetlabs_spec_helper/rake_tasks.rb

### DIFF
--- a/moduleroot/Rakefile
+++ b/moduleroot/Rakefile
@@ -1,7 +1,4 @@
 require 'puppetlabs_spec_helper/rake_tasks'
-require 'puppet-lint/tasks/puppet-lint'
-require 'puppet-syntax/tasks/puppet-syntax'
-require 'metadata-json-lint/rake_task'
 require 'puppet_blacksmith/rake_tasks'
 require 'voxpupuli/release/rake_tasks'
 require 'rubocop/rake_task'


rnelson0:
For lack of ability to test this myself right now, got a link to a module with this trimmed file that tests properly?

alexharv074:
How about `lib/puppetlabs_spec_helper/rake_tasks.rb` at [https://github.com/puppetlabs/puppetlabs_spec_helper/blob/master/lib/puppetlabs_spec_helper/rake_tasks.rb#L6-L11](lines 6-11), [376](https://github.com/puppetlabs/puppetlabs_spec_helper/blob/master/lib/puppetlabs_spec_helper/rake_tasks.rb#L376) and [393](https://github.com/puppetlabs/puppetlabs_spec_helper/blob/master/lib/puppetlabs_spec_helper/rake_tasks.rb#L393).

I tested by removing all those lines and confirming that I had 23 rake tasks with or without them in `rake -T`.  I don't know of a module that has this config, other than one I'm not quite ready to push.  Will this do?

alexharv074:
The latest version of `lib/puppetlabs_spec_helper/rake_tasks.rb` at [line 3](https://github.com/puppetlabs/puppetlabs_spec_helper/blob/master/lib/puppetlabs_spec_helper/rake_tasks.rb#L3) also requires rubocop, but seems like that hasn't been released unfortunately.

bastelfreak:
I just tested this on puppet-zabbix, I get the same amount of rake tasks after applying your patch (42).

PR #149, opened 2016-05-08
add missing encoding to file header

bastelfreak:
Rubocop complaints about this if the cop isn't disabled.

### DIFF
--- a/moduleroot/spec/classes/coverage_spec.rb
+++ b/moduleroot/spec/classes/coverage_spec.rb
@@ -1,3 +1,4 @@
+# encoding: utf-8
 require 'rspec-puppet'
 
 at_exit { RSpec::Puppet::Coverage.report! }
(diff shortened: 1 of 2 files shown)


rnelson0:
...isn't the encoding done on the file, not by comments at the top of it? Does that actually force a non-utf8 file to be utf8?

bastelfreak:
I'm not sure how correct this answer is: http://stackoverflow.com/a/8879359

rnelson0:
http://stackoverflow.com/a/8796676 May be more relevant. Looks like this was only ever needed with Ruby 1.9. I know we are testing with that but since we don't have the cop enabled yet, I think it's premature. Hopefully 1.9 gets out of the picture soon anyway.

jyaworski:
Should we be explicit, though? Like with vim modelines?

rnelson0:
I am generally against pre-mature optimizations :)

daenney:
I'm closing this for now. I don't think it's needed and I haven't seen any issues with it in the past 5 years or so. If it crops up we can deal with it then 😄.

PR #153, opened 2016-05-10
This cop is useless for Puppet

jyaworski:
@rnelson0 @bastelfreak


rnelson0:
Definitely not helpful. Was this biting us, though? I had not seen failures for it before.

jyaworski:
@rnelson0 yes, it was in the test PR I had with collectd. The first line of every file would fail due to the format.

rnelson0:
Is that new with the latest version of rubocop or something?

jyaworski:
@rnelson0 rubocop-rspec.

PR #154, opened 2016-05-10
remove rubocop from our gemlist

bastelfreak:
we have rspec-rubocop now, is has rubocop as a dependency
https://rubygems.org/gems/rubocop-rspec


igalic:
we'd have to pin rubocop-rspec then, or else we're chasing a moving target again

bastelfreak:
@igalic done

PR #156, opened 2016-05-12
Add vim modelines to everything. Fixes #150

### LINKED ISSUE #150: Add vim modelines to files
We should add vim modelines to our files, because we all use vim.

jyaworski:
@bastelfreak @rnelson0


bastelfreak:
I'm fine with the modelines. But you're also changing comments. Please rename the commit the mention that or even more awesome: split that in a separate commit. Git can easily do that with interactively staging.

PR #157, opened 2016-05-12
Add debian 6 to acceptance testing, use puppet boxes

jyaworski:
@rnelson0 @bastelfreak


bastelfreak:
why adding debian 6? this is EOL since a long long time. even their LTS support is now dropped. We're currently trying to get rid of ruby1.8, testing on debian6 would be a step backwards IMO.

jyaworski:
@bastelfreak Puppet itself is still testing against it (per their vagrant boxes).

bastelfreak:
I know, but still, long term EOL, I am against that. Does anybody else have opinions here? @igalic or @nibalizer?

igalic:
i'm not sure i agree with this…

PR #159, opened 2016-05-20
Rspec length

jyaworski:


bastelfreak:
Could you add a comment why you disable this?

jyaworski:
@bastelfreak done.

PR #162, opened 2016-05-21
bump ruby to 2.3.1, don't ignore fails

bastelfreak:
Ruby 2.3.1 is getting serious in the puppet world. Puppet will migrate their AIO packages to 2.3.1 in the near future so we should take testing on that version serious.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ build/
 ## Environment normalisation:
 /.bundle/
 /vendor
+/.vendor
 /lib/bundler/man/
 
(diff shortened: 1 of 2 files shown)


rnelson0:
Should we test using 2.3 instead of specifying the patch level? We do that for 2.2.

jyaworski:
@rnelson0 that depends on if rvm got its act together with the alias.

bastelfreak:
I tested it, doesn't work. travis requires the patch level for ruby 2.3.

PR #169, opened 2016-05-27
Default facts

jyaworski:


moduleroot/spec/spec_helper.rb
```diff
@@ -1,6 +1,14 @@
 require 'puppetlabs_spec_helper/module_spec_helper'
 require 'rspec-puppet-facts'
 include RspecPuppetFacts
+
```
rnelson0: Wrap this in a test whether default_facts has contents, in case someone overrides it in a module's `.sync.yml`.

jyaworski: This will read the default_facts in .sync.yml already.

rnelson0: Yes, and if default_facts is empty?

jyaworski: It would only be empty if someone specifically overrode it in .sync.yml to be a blank hash. I added a default value in config_defaults.

jyaworski: Even so, I added an unless to it.

PR #176, opened 2016-06-06
enable more cops for better code quality

bastelfreak:
there is a long list of cops available, we honour already many of them
even if we don't explicitly enable them. This commit changes that. All
settings were tested on the dhcp and/or zabbix module.

### DIFF
--- a/moduleroot/.rubocop.yml
+++ b/moduleroot/.rubocop.yml
@@ -267,7 +267,7 @@ Style/DeprecatedHashMethods:
   Enabled: true
 
 Style/DotPosition:
-  Enabled: true
+  EnforcedStyle: trailing
 
 Style/DoubleNegation:
   Enabled: true
--- a/moduleroot/.rubocop.yml
+++ b/moduleroot/.rubocop.yml
@@ -452,7 +452,7 @@ Style/RescueModifier:
 
 # based on https://github.com/voxpupuli/modulesync_config/issues/168
 Style/RegexpLiteral:
-	EnforcedStyle: percent_r
+  EnforcedStyle: percent_r
   Enabled: true
 
 Lint/UnderscorePrefixedVariableName:
--- a/moduleroot/.rubocop.yml
+++ b/moduleroot/.rubocop.yml
@@ -13,22 +13,53 @@ AllCops:
     - <%= x %>
     <%- end -%>
     <%- end -%>
+Lint/ConditionPosition:
+  Enabled: true
 
-# Configuration parameters: AllowURI, URISchemes.
-Metrics/LineLength:
-  Max: 328
+Lint/ElseLayout:
+  Enabled: true
 
-# 'Complexity' is very relative
-Metrics/PerceivedComplexity:
-  Enabled: false
+Lint/UnreachableCode:
+  Enabled: true
 
-# 'Complexity' is very relative
-Metrics/CyclomaticComplexity:
-  Enabled: false
+Lint/UselessComparison:
+  Enabled: true
 
-# 'Complexity' is very relative
-Metrics/AbcSize:
-  Enabled: false
+Lint/EnsureReturn:
+  Enabled: true
+
+Lint/HandleExceptions:
+  Enabled: true
+
+Lint/LiteralInCondition:
+  Enabled: true
+
+Lint/ShadowingOuterLocalVariable:
+  Enabled: true
+
+Lint/LiteralInInterpolation:
+  Enabled: true
+
+Style/HashSyntax:
+  Enabled: true
+
+Style/RedundantReturn:
+  Enabled: true
+
+Lint/AmbiguousOperator:
+  Enabled: true
+
+Lint/AssignmentInCondition:
+  Enabled: true
+
+Style/SpaceBeforeComment:
+  Enabled: true
+
+Style/AndOr:
+  Enabled: true
+
+Style/RedundantSelf:
+  Enabled: true
 
 # Method length is not necessarily an indicator of code quality
 Metrics/MethodLength:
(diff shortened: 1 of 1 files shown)


moduleroot/.rubocop.yml
```diff
@@ -13,22 +13,53 @@ AllCops:
...
+  Enabled: true
+
+Style/RedundantReturn:
+  Enabled: true
+
+Lint/AmbiguousOperator:
+  Enabled: true
+
+Lint/AssignmentInCondition:
+  Enabled: true
```
igalic: we use this in all types & providers code :D

moduleroot/.rubocop.yml
```diff
@@ -38,50 +69,437 @@ Metrics/MethodLength:
...
+
+Lint/DefEndAlignment:
+  Enabled: true
+
+Lint/EndAlignment:
+  Enabled: true
+
+Lint/DeprecatedClassMethods:
+  Enabled: true
```
igalic: cool

moduleroot/.rubocop.yml
```diff
@@ -38,50 +69,437 @@ Metrics/MethodLength:
...
+
+Lint/Loop:
+  Enabled: true
```
igalic: what does this do?

moduleroot/.rubocop.yml
```diff
@@ -38,50 +69,437 @@ Metrics/MethodLength:
...
+
+Lint/UnusedBlockArgument:
+  Enabled: true
+
+Lint/UnusedMethodArgument:
+  Enabled: true
+
+Lint/UselessAccessModifier:
+  Enabled: true
```
igalic: what does this do?

moduleroot/.rubocop.yml
```diff
@@ -38,50 +69,437 @@ Metrics/MethodLength:
...
+
+Style/AlignHash:
+  Enabled: true
+
+Style/AlignParameters:
+  Enabled: true
+
+Metrics/BlockNesting:
+  Enabled: true
```
igalic: :+1:

moduleroot/.rubocop.yml
```diff
@@ -38,50 +69,437 @@ Metrics/MethodLength:
...
+
+Style/AsciiComments:
+  Enabled: true
```
igalic: lolsob yes :(

moduleroot/.rubocop.yml
```diff
@@ -38,50 +69,437 @@ Metrics/MethodLength:
...
+Style/DefWithParentheses:
   Enabled: true
 
-Style/BlockDelimiters:
+Style/DeprecatedHashMethods:
   Enabled: true
 
-Lint/AmbiguousRegexpLiteral:
+Style/DotPosition:
+  Enabled: true
```
igalic: where does this require us to put the . ?

moduleroot/.rubocop.yml
```diff
@@ -38,50 +69,437 @@ Metrics/MethodLength:
...
+
+Style/DoubleNegation:
+  Enabled: true
```
igalic: i hope this exempts the `!!foo` case to convert any type into boolean

igalic: if it does, it basically takes a way a lot of my work during reviews :D

moduleroot/.rubocop.yml
```diff
@@ -38,50 +69,437 @@ Metrics/MethodLength:
...
+
+Style/TrailingCommaInLiteral:
+  Enabled: true
+
+Style/GlobalVars:
+  Enabled: true
+
+Style/GuardClause:
+  Enabled: true
```
igalic: hm?

moduleroot/.rubocop.yml
```diff
@@ -38,50 +69,437 @@ Metrics/MethodLength:
...
+
+Style/Proc:
+  Enabled: true
+
+Style/RaiseArgs:
+  Enabled: true
+
+Style/RedundantBegin:
+  Enabled: true
```
igalic: it's really hard to tell what a lot of these do…
Do they permit OneLineConditionals or not? etc…

jyaworski:
I approve of everything here.

nibalizer:
I have no opinon

moduleroot/.rubocop.yml
```diff
@@ -160,7 +160,7 @@ Style/ClassAndModuleCamelCase:
   Enabled: true
 
 Style/ClassAndModuleChildren:
-  Enabled: true
+  Enabled: false
```
rski: I think i saw you mention it in the irc, but why disable it? It seems like its suggestions are sane, no?

bastelfreak: I had some issues where I wasn't able to fix the code. The idea was that it is a false positive or I'm to stupid to fix it. My ruby foo isn't the best so it maybe was my issue. I'm fine with re-enabling it.

rski: It does look like a false positive. I'd say merge it

moduleroot/.rubocop.yml
bastelfreak: Style/RedundantBegin: useless use of a begin/end block - for example through a complete method:
```ruby
def redundant
  begin
    ala
    bala
  rescue StandardError => e
    something
  end
end

def preferred
  ala
  bala
rescue StandardError => e
  something
end
```

----

Style/RaiseArgs checks how stuff is passed to a raise/fail blog.

---

Style/Proc: This cop checks for uses of Proc.new where Kernel#proc would be more appropriate.

---

Style/SelfAssignment enforces stuff like:
```ruby
a += 1337
```
where this is prohibited:
```ruby
a  = a+ 1337
```

---

Style/RedundantException checks for a redundant block in a raise/fail section

----

(I copied most of the explanations from the rubocop docs)

moduleroot/.rubocop.yml
bastelfreak: GuardClause is pretty awesome I think:
```ruby
# bad
def test
  if something
    work
    work
    work
  end
end

# good
def test
  return unless something
  work
  work
  work
end
```

moduleroot/.rubocop.yml
bastelfreak: throws an error if you set an access modifier but never use it:
```ruby
class Foo
  private # This is useless

  def self.some_method
  end
end
```

moduleroot/.rubocop.yml
bastelfreak: from the docs:
'Use `Kernel#loop` with `break` rather than `begin/end/until`(or `while`).'

moduleroot/.rubocop.yml
bastelfreak: Puppet best practice differ a bit from ruby best practice. As far as I know this is best practice in ruby and we should try to adopt it. We can still disable it again if it doesn't work out.

moduleroot/.rubocop.yml
igalic: i'm not sure i've ever seen that, but, okay.

moduleroot/.rubocop.yml
igalic: :+1:

moduleroot/.rubocop.yml
igalic: again, this would take away a lot of my review work :D

PR #183, opened 2016-06-24
don't explicitly include puppet-lint

bastelfreak:
we've multiple gems as puppet-lint plugins that all requier puppet-lint

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -23,8 +23,6 @@ Gemfile:
         git: https://github.com/rodjek/rspec-puppet.git
       - gem: rspec-puppet-facts
       - gem: rspec-puppet-utils
-      - gem: puppet-lint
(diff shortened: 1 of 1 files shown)


bastelfreak:
We should discuss if we want to pull it from rubygems instead of git directly or if we want it as an implicit dep. based on a discussion at https://github.com/voxpupuli/puppet-corosync/pull/292

alexjfisher:
LGTM.  Personally, I'd say go with the released version.

rnelson0:
PLEASE do not pull from head, or sometime in the future modules will pull 3.0.0 directly but the checks will be looking for `'>= 1.0.0', '< 3.0.0'` and the world will be broken again. 100% on board with this change as is to prevent that from happening. Again.

PR #185, opened 2016-06-27
Fix github templates

roidelapluie:
- Keep modules simple
- Do not scare people with bullet points
- Stop making pull requests unreadable
- Keep contributing.md

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

### DIFF
--- a/moduleroot/.github/ISSUE_TEMPLATE.md
+++ b/moduleroot/.github/ISSUE_TEMPLATE.md
@@ -1,25 +1,22 @@
-Please check the following items before submitting an issue -- thank you!
-
-Note that this project is released with a Contributor Code of Conduct.
-By participating in this project you agree to abide by its terms.
-[Contributor Code of Conduct](https://voxpupuli.org/coc/).
-
-- [ ] There is no existing issue or PR that addresses this problem
-
-Optional, but makes our lives much easier:
-
-- [ ] The issue affects the latest release of this module at the time of
-  submission
-
-- - -
+<!--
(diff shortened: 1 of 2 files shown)


roidelapluie:
scrnshot

CONTRIBUTING.md is already there and less intrusive -- while still visible enough to drag attention

bastelfreak:
I'm against this change. I don't think that many people read the CONTRIBUTING.md before they submit a PR. We maybe should reorganize the templates, but removing them is IMO the wrong way.

igalic:
I absolutely agree with @roidelapluie here: when i created a new pull request after we introduced .github i was *utterly* confused — and i'm not quite a newbie anymore.

The Contributing.md is a good place to stuff general docs. If we want a step-by-step how-to do a good PR, we can write that as a blog post and point newbies to. Same, IMO, for contributing — if someone has a question that's answered there, you can point them to it. I see no reason to modify someone's pr, it's *highly* intrusive!

We've already (mostly) gotten rid of the requirement for squashing / rebasing because github can do that for us by now, so i really see this as a step back.

juniorsysadmin:
+1 The slabs of text have been freaking me out

rski:
I agree on the PR template, because it confused me too.

I disagree on the issue template though. Its points are pretty much things that will be asked anyway by the person who decides to pick up the issue and fix it.

roidelapluie:
Okay, what about this?

rski:
+1

igalic:
:+1:

PR #190, opened 2016-07-03
Pin rest-client gem to avoid breakage on Ruby 1.9x

juniorsysadmin:
The rest-client gem is required by puppet-blacksmith, but 2.x of the gem requires Ruby 2.0 or above

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -35,6 +35,10 @@ Gemfile:
       - gem: metadata-json-lint
       - gem: puppet-blacksmith
         git: https://github.com/voxpupuli/puppet-blacksmith.git
+        # puppet-blacksmith requires rest-client, but rest-client 2.x requires
+        # Ruby 2.0 or above
+      - gem: rest-client
+        version: '~> 1.8'
       - gem: voxpupuli-release
         git: https://github.com/voxpupuli/voxpupuli-release-gem.git
       - gem: puppet-strings


bastelfreak:
in the past, pinning often resulted into bad errors and dependency cycles. What about moving blacksmith and puppet-release into a new group, called release. We can exclude that on travis tests.

* do we need the gems if we want to release something via travis or only locally?
* can't we drop ruby1.9?

igalic:
no. we cannot drop ruby 1.9
the way we pinned things in the past was to pin it (temporarily) based on the ruby version.

config_defaults.yml
```diff
@@ -35,6 +35,10 @@ Gemfile:
       - gem: metadata-json-lint
       - gem: puppet-blacksmith
         git: https://github.com/voxpupuli/puppet-blacksmith.git
+        # puppet-blacksmith requires rest-client, but rest-client 2.x requires
+        # Ruby 2.0 or above
+      - gem: rest-client
+        version: '~> 1.8'
```
igalic: will this work?
wouldn't we have to pin this in puppet-blacksmith for it to work?

juniorsysadmin: puppet-blacksmith doesn't seem to care about the version

igalic: then let's pin it there, and open an issue that it needs fixing.

alexjfisher:
https://github.com/voxpupuli/puppet-blacksmith/pull/35 pins it in puppet-blacksmith

PR #192, opened 2016-07-03
Disable Style/TrailingCommaInLiteral cop

juniorsysadmin:
The spec_overrides code in spec_helper.rb always puts a comma on each line.

### DIFF
--- a/moduleroot/.rubocop.yml
+++ b/moduleroot/.rubocop.yml
@@ -322,7 +322,7 @@ Style/TrailingCommaInArguments:
   Enabled: true
 
 Style/TrailingCommaInLiteral:
-  Enabled: true
+  Enabled: false
 
 Style/GlobalVars:
   Enabled: true


juniorsysadmin:
This cop is not great for Pull Request newbies either in my opinion

juniorsysadmin:
There possibly needs to be a review of the cops that we have. Are all of them enforcing great code, or are they potentially causing unnecessary barriers to a Pull Request getting merged? Or have they already been reviewed already and there is no need for this question.

bastelfreak:
We introduced a lot of them in the past weeks, they should have been reviewed, but more reviewers are always great. There is always a trade off between a consistent style across all of our repos and the barrier for a beginner. In my opinion, the consistent style is important. Most of the cops we have can be handled by the autofix option on rubocop, so the barrier shouldn't be that high.

jyaworski:
@juniorsysadmin is this still needed? Can you rebase?

PR #204, opened 2016-08-01
fix rubocop-rspec

bastelfreak:
rubocop-rspec dropped support for ruby2 and ruby21 here:
https://github.com/nevir/rubocop-rspec/commit/8a68286c31afef6a777b5b4dc65a32889df881d4

This got today released as version 1.5.2.

This commit only installs rubocop-rspec if we're on ruby2.3 or newer. We
tested it successfully here:
https://github.com/voxpupuli/puppet-community_kickstarts/pull/15
https://travis-ci.org/voxpupuli/puppet-community_kickstarts/builds/148977207

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -41,7 +41,7 @@ Gemfile:
       - gem: rubocop-rspec
         version: '~> 1.5'
         ruby-operator: '>='
-        ruby-version: '2.0.0'
+        ruby-version: '2.3.0'
       - gem: json_pure
         version: '<= 2.0.1'
         ruby-operator: '<'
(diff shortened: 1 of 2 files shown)


config_defaults.yml
```diff
@@ -41,7 +41,7 @@ Gemfile:
       - gem: rubocop-rspec
         version: '~> 1.5'
         ruby-operator: '>='
```
igalic: i didn't see this… that feels… bad :P

config_defaults.yml
```diff
@@ -41,7 +41,7 @@ Gemfile:
...
-        ruby-version: '2.0.0'
+        ruby-version: '2.3.0'
```
igalic: i thought we already did that elsewhere…?
i.e.: we only run rspec against the latest ruby/puppet anyway

igalic: basically, this now aligns our travis test matrix with our Gemfile and Rakefile

bastelfreak: yep in the Rakefile. But we still have to do it here too :(

igalic: yeah, i know, all i'm saying is that we're now consistent.

PR #206, opened 2016-08-05
read default_module_facts.yml if it exists

3flex:
Currently it's reading `default_facts.yml` twice if `default_module_facts.yml` exists.


bastelfreak:
oh sorry for introducing this :(

PR #207, opened 2016-08-05
make loading default facts more robust

3flex:
* Merge instead of attempting to append - means default_module_facts can override default_facts
* Make the loading of the files independent of the current working directory
* YAML.read_file doesn't appear to be a valid method - replaced with YAML.load

Tested on Ruby 1.9.3, 2.1, 2.2 and 2.3.1. See results at https://travis-ci.org/jfryman/puppet-nginx/builds/150137623, which tested this commit: https://github.com/jfryman/puppet-nginx/commit/c529a148e5e6436c578fd52527f179ddfdc5fa01

### DIFF
--- a/moduleroot/spec/spec_helper.rb
+++ b/moduleroot/spec/spec_helper.rb
@@ -7,8 +7,8 @@ RSpec.configure do |c|
     puppetversion: Puppet.version,
     facterversion: Facter.version
   }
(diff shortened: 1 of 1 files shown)


3flex:
Regarding the second point (making the loading files independent of current working directory), the files were never loaded when running `bundle exec rake test` without the change to checking for existence with `File.expand_path('../default_facts.yml', __FILE__)`

bastelfreak:
whups, thanks for fixing!

PR #208, opened 2016-08-07
Ruby2.4 support + minor fix

bastelfreak:
Would be totally awesome if github could show here the text of all commits.

```
commit e5bb33c3e5eb13c999ebded2ba6474c0c487dd7c
Author: Tim Meusel <tim@bastelfreak.de>
Date:   Sun Aug 7 12:15:38 2016 +0200

    use capitalized bools
    
    We already used it a few times in the rubocop config. this patch makes
    it now consistent.
    (We could also go with lowercase here, I just prefer capitalized)


commit 13944658ab4a7ec7d61c006423d0ab25d4cb98f9
Author: Tim Meusel <tim@bastelfreak.de>
Date:   Sun Aug 7 12:10:43 2016 +0200

    disable RSpec/NamedSubject for now
    
    this is a complicated cop that isn't easy to fix. It got introduced in
    the last rubocop-rspec release. We also use it in a lot of modules so
    fixing is time consuming and breaks our modulesync. We already disabled
    the cop in https://github.com/voxpupuli/puppet-nodejs/pull/240 which
    created the expected result.
    lot of modules


commit b8141b95717d38f0d2051d089cb87483f4369e34
Author: Tim Meusel <tim@bastelfreak.de>
Date:   Sun Aug 7 12:00:13 2016 +0200

    bump rubocop-rspec to 1.6
    
    We already test with version 1.6 and updated our configs to reflect the
    changes. There are a few new cops that could break on older
    rubocop-rspec 1.5, so we should bump the requoired version. Initial 1.6
    support came in https://github.com/voxpupuli/modulesync_config/pull/204


commit 5c1ec77c6f4e016d3019d43308b95c2cf9064585
Author: Tim Meusel <tim@bastelfreak.de>
Date:   Sun Aug 7 11:48:08 2016 +0200

    add testing on ruby2.4.0
    
    this is becomming the new Ruby version in the near future. The rolling
    release distributions will soon update to it, so we should already test
    against it. It is still under development, so we enable allow_failures
    for it. Tested in https://github.com/voxpupuli/puppet-gluster/pull/60


commit 072f5d6112825d2a87d71afa7a93af3b11562dbd
Author: Tim Meusel <tim@bastelfreak.de>
Date:   Sun Aug 7 11:44:32 2016 +0200

    remove uesless env attribute from allow_failures
    
    this isn't needed here. We specify env + rvm version in our build
    matrix. allow_failures only needs to list the rvm version. This is
    tested in https://github.com/voxpupuli/puppet-gluster/pull/60 and
    documented at
    https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail
```

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -17,6 +17,10 @@
     env: PUPPET_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=rubocop
   - rvm: 2.3.1
     env: PUPPET_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=test
+  - rvm: 2.4.0-preview1
+    env: PUPPET_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=test
+  allow_failures:
+    - rvm: 2.4.0-preview1
 Gemfile:
   required:
     ':test':
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -43,7 +43,7 @@ Gemfile:
       - gem: puppet-strings
         git: https://github.com/puppetlabs/puppetlabs-strings.git
       - gem: rubocop-rspec
-        version: '~> 1.5'
+        version: '~> 1.6'
         ruby-operator: '>='
         ruby-version: '2.3.0'
       - gem: json_pure
--- a/moduleroot/.rubocop.yml
+++ b/moduleroot/.rubocop.yml
@@ -14,476 +14,476 @@ AllCops:
     <%- end -%>
     <%- end -%>
 Lint/ConditionPosition:
-  Enabled: true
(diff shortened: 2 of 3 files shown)


daenney:
We had a discussion about testing on Ruby 2.4-preview on IRC. The few people involved voiced the opinion that it's likely not helpful to us to test on 2.4-preview, especially since the language hasn't settled and any bugs surfacing here are more likely to be issues in Puppet + Ruby 2.4 but @bastelfreak feels strongly enough about it that we're merging it anyway. It doesn't hurt us since it's in the `allowed_failures` not blocking PRs but for those that care enough the data will be available in Travis.

PR #209, opened 2016-08-15
(Re-)enable Style/ClosingParenthesisIndentation cop

3flex:
It was enabled in ed0f3ce5 then apparently inadvertently disabled in d7e9b20c


bastelfreak:
ups, sorry for removing it :(

PR #214, opened 2016-08-23
pin rubocop and rubocop-rspec depending on Ruby version

3flex:
This table shows the last supported versions of rubocop and rubocop-rspec against each Ruby version supported by the default modulesync config.

Ruby version|rubocop|rubocop-rspec
---|---|---
1.9.3|0.41.2|1.5.0
2.1|Supported|1.5.1
2.2|Supported|Supported
2.3|Supported|Supported

This PR sets the versions in line with the above, except for Ruby 2.1 which gets rubocop-rspec 1.5.0 even though it could get 1.5.1. There's nothing interesting in 1.5.1 so it was left out to reduce complexity of the config.

To summarize:
* Ruby 1.9.3 gets rubocop 0.41.2 and rubocop-rspec 1.5.0
* Ruby 2.1 gets latest rubocop and rubocop-rspec 1.5.0
* Ruby 2.2 and higher gets latest rubocop and rubocop-rspec 1.6.x

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -47,7 +47,15 @@ Gemfile:
       - gem: rubocop-rspec
         version: '~> 1.6'
         ruby-operator: '>='
-        ruby-version: '2.3.0'
+        ruby-version: '2.2.0'
+      - gem: rubocop-rspec
+        version: '1.5.0'
+        ruby-operator: '<'
+        ruby-version: '2.2.0'
+      - gem: rubocop
+        version: '0.41.2'
+        ruby-operator: '<'
+        ruby-version: '2.0.0'
       - gem: json_pure
         version: '<= 2.0.1'
         ruby-operator: '<'


bastelfreak:
I'm not sure if we should merge this. We only run rubocop in travis on ruby231. Our Gemfile and Rakefile are tuned for that and it works perfectly well. This PR adds a bunch of dependencies, not sure if that is a good idea, and we even don't use them. Also I don't want to encourage people to use old and unsupported ruby versions.

---

 I would like to have some input here from other collaborators, but I tend to not merge it.

3flex:
Sorry, I forgot to mention what triggered the need for this... puppetlabs_spec_helper as of 1.2.0 depends on rubocop and rubocop-rspec, so the gems will now be installed for all Ruby versions in the Travis matrix, not just Ruby 2.3. This fails when that's attempted for unsupported Ruby versions.

Also note that rubocop-rspec is already being installed as a dependency in this gem group, which already pulls in rubocop. So the effect isn't adding any new dependencies, just making sure the correct version is installed.

puppetlabs_spec_helper was just released today, so you're not seeing mass failures in Travis yet, but there will be on the next sync. But here's one example that's now failing: https://travis-ci.org/voxpupuli/puppet-corosync/builds/154550883

bastelfreak:
ah yeah. We discussed this on IRC. There will be an update for PSH tomorrow or in the next days, so we should wait with this for at least a few hours I guess.

3flex:
I see the upstream issue: https://tickets.puppetlabs.com/browse/MODULES-3776

3flex:
PSH 1.2.0 was yanked, so this is no longer an issue.

PR #219, opened 2016-08-30
Create a DEPLOY_TO_FORGE constraint

roidelapluie:
This adds a more visible condition to see whch job goes to forge. It is
also easier to avoid duplicates.

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>


bastelfreak:
tested on https://github.com/voxpupuli/puppet-graphite_powershell

PR #220, opened 2016-08-30
Make .gitignore fixtures more specific

alexjfisher:
It's quite valid to have directories in spec/fixtures that should be
part of what's committed.
eg in `puppet-archive` there is a `spec/fixtures/checksum` directory.

`spec/fixtures/manifests` and `spec/fixtures/modules` get populated by
`spec_prep` on the other hand.


bastelfreak:
tested in https://github.com/voxpupuli/puppet-archive/pull/207

PR #222, opened 2016-08-31
update ruby patch releases

bastelfreak:
this got tested at
https://travis-ci.org/bastelfreak/puppet-mirror/jobs/156647703


igalic:
i'm confused.
can you elaborate (in your commit message) why rvm isn't handling this itself?
why do we have to manually bump versions??

bastelfreak:
RVM does not provide all patch releases by itself. Using `2.1` will use the latest availavle 2.1.X release that rvm serves. Using `2.1.9` will use the this specific version. RVM uses http://rubies.travis-ci.org/ as source if it doesn't find the version on rvm.org

bastelfreak:
We already had to do that for ruby2.3.1

igalic:
this sucks :(

PR #227, opened 2016-09-16
Fix for latest `puppet-strings`

alexjfisher:
https://github.com/puppetlabs/puppet-strings/pull/98 is massive, but I
think this is the most important thing we need to change right now.

Without this fix, we currently can't run *any* rake tasks.

### DIFF
--- a/moduleroot/Rakefile
+++ b/moduleroot/Rakefile
@@ -1,7 +1,7 @@
 require 'puppetlabs_spec_helper/rake_tasks'
 require 'puppet_blacksmith/rake_tasks'
 require 'voxpupuli/release/rake_tasks'
-require 'puppet-strings/rake_tasks'
+require 'puppet-strings/tasks'
 
 if RUBY_VERSION >= '2.3.0'
   require 'rubocop/rake_task'


alexjfisher:
See https://github.com/puppetlabs/puppet-strings/pull/98/files#diff-04c6e90faac2675aa89e2176d2eec7d8R336

alexjfisher:
Alternatively, if puppet-strings is about to enter a period of instability, we could fix the Gemfile instead and fix on the commit before the breaking change?

rski:
@alexjfisher yeah that sounds like a good compromise for now, there might be other stuff breaking with such a big patch

alexjfisher:
@rski Ok.  Let's hold off merging this until a consensus is formed.

Pinging the PMC nominees! ;) @bastelfreak @nibalizer @hunner @bbriggs @eputnam @dhoppe

bbriggs:
Do you have any indication that puppet-strings is about to become unstable for a while?

alexjfisher:
@bbriggs I think the opposite https://github.com/puppetlabs/puppet-strings/pull/98#issuecomment-247672603

bbriggs:
I mean, other than the fact that the PR is massive and the likelihood we'll keep finding bugs for the next couple months is high ;)

alexjfisher:
Do we have any automated builds that actually run any of the puppet-string tasks?  Dunno.

The immediate problem is that *all* builds are failing with
`
LoadError: cannot load such file -- puppet-strings/rake_tasks
`

bastelfreak:
I think the only module that uses puppet-strings is our new php one? I would prefer it if we pin it to the last release. Using master branch in generally a bad idea, we should work with releases.

dhoppe:
The Nginx module also uses puppet-strings.

alexjfisher:
@bastelfreak I agree we should move to working with releases.  But how about we do that starting with 1.0.0 which should land in the next couple of weeks?
0.4.0 is from March, even ignoring  https://github.com/puppetlabs/puppet-strings/pull/98, there have been a fair few commits since then.  Maybe some projects are relying on them??

bbriggs:
Step 1 in this whole process should be identifying our repos that rely on
puppet-strings to see the scope of the change. However, since this is a
recent break (and not on a release), pinning to a release/tag seems to be
the reasonable and straightforward fix.

PR #229, opened 2016-09-24
add Ubuntu 16.04 as default nodeset

3flex:
Using Ubuntu over CentOS might be contentious... not sure what's generally used more often.

But since these are typically used for automated testing Ubuntu seemed as good a choice as any.

### DIFF
--- a/moduleroot/spec/acceptance/nodesets/default.yml
+++ b/moduleroot/spec/acceptance/nodesets/default.yml
@@ -1,0 +1,12 @@
+---
+HOSTS:
+  ubuntu-server-1604-x64:
+    roles:
+      - master
+    platform: ubuntu-16.04-amd64
+    box: puppetlabs/ubuntu-16.04-64-nocm
+    hypervisor: vagrant
+CONFIG:
+  type: foss
+...
+# vim: syntax=yaml


3flex:
Note that this doesn't conflict with #884. If both get merged, then someone running `bundle exec rake beaker` on their local machine would find acceptance tests run on Vagrant, using this host file for the host's configuration. #884 is only used in Travis.

dhoppe:
As far as I know, we do not provide a default nodeset for our modules. Travis CI can be configured to test multiple operating systems and the user should be able to choose as he likes.

3flex:
This is for people running acceptance tests on their own machine(s), so they don't have to manually set BEAKER_set which is an extra step that's unnecessary if a default nodeset is provided. It reduces friction for new contributors.

I see your argument for not including it, so you can close if you prefer, or wait for feedback from others.

wyardley:
@3flex: FWIW, rake will already give various options, like:
```
rake beaker:centos-66-x64
rake beaker:centos-66-x64-pe
[...]
rake beaker:ssh:centos-66-x64
[...]
rake beaker:ubuntu-server-1604-x64
```
and so on. So it is possible to build on various platforms without setting `BEAKER_x` directly, however, `rake beaker` and `rake acceptance` will fail, I think, without a default nodeset. Is there an easy way for projects to set a per-project default nodeset without messing up the modulesync process?

One other thing we could probably add that might help would be more info in the CONTRIBUTING file.

wyardley:
ps - 16.04 is not yet working for me, so in any event, let's fix that problem first.
I think there's only -pc1. changing type to 'aio' from 'foss' didn't help, looking into what will fix it...

```
/usr/local/lib/ruby/gems/2.3.0/gems/beaker-2.51.0/lib/beaker/host.rb:351:in `exec': Host 'ubuntu-server-1604-x64' exited with 8 running: (Beaker::Host::CommandFailure)
 wget -O /tmp/puppet.deb http://apt.puppetlabs.com/puppetlabs-release-xenial.deb
Last 10 lines of output were:
	--2016-10-27 10:21:08--  http://apt.puppetlabs.com/puppetlabs-release-xenial.deb
```

igalic:
@3flex why's this not simply a symlink? or will that… make windows sad?

juniorsysadmin:
>  Or it could be added to .sync.yml to allow the module to choose the default nodeset, and enhance msync to copy it to nodesets/default.yml on a module sync

That sounds like the best option, but I'm okay with no default or a default

PR #232, opened 2016-10-07
add a note about rubocop checks, since they're enforced

wyardley:
Since rubocop checks are enforced in Travis, there should be a note about this.


wyardley:
I also changed the version of CentOS specified in the example for beaker to match the current CentOS 6 node in `spec/acceptance/nodesets`. I think the docs are still wrong about rake acceptance running on an Ubuntu 12 box with no arguments, since there doesn't seem to be a default.yml in that directory.

PR #237, opened 2016-10-11
Use new key line over linenumber.

traylenator:
Fixes where you can run into  the error below.

The puppet-lint deprecation

http://github.com/rodjek/puppet-lint/pull/540

has been merged into master.

According to

https://github.com/rodjek/puppet-lint/issues/539

it was meant to be added at version 3 but is now merged
with 2.0.2.

Thanks @Feandil for diagnosis.

```
rake aborted!
KeyError: key{linenumber} not found
/afs/cern.ch/user/s/straylen/.gem/ruby/bundler/gems/puppet-lint-32e93d43a3bf/lib/puppet-lint.rb:102:in
`%'
/afs/cern.ch/user/s/straylen/.gem/ruby/bundler/gems/puppet-lint-32e93d43a3bf/lib/puppet-lint.rb:102:in
`format_message'
/afs/cern.ch/user/s/straylen/.gem/ruby/bundler/gems/puppet-lint-32e93d43a3bf/lib/puppet-lint.rb:136:in
`block in report'
```

### DIFF
--- a/moduleroot/Rakefile
+++ b/moduleroot/Rakefile
@@ -12,7 +12,7 @@ if RUBY_VERSION >= '2.3.0'
   end
 end
 
-PuppetLint.configuration.log_format = '%{path}:%{linenumber}:%{check}:%{KIND}:%{message}'
+PuppetLint.configuration.log_format = '%{path}:%{line}:%{check}:%{KIND}:%{message}'
 PuppetLint.configuration.fail_on_warnings = true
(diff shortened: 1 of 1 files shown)


rski:
I merged this because line has been present for a long time so it'll work for any version of puppet-lint >1.0.
I wonder though, do we track the master branch of puppet-lint? Because backwards incompatible stuff could be merged at any point in preparation for the next major release.
@bastelfreak any thoughts?

Issue #80, opened 2016-02-27
Add new .travis.yml section for beaker-docker tests


petems:
Although we have PCCI for acceptance testing, since it's fairly easy to setup Beaker to docker tests in Travis CI, might be an idea to add that as a configurable section to the .travis.yml for modules?

``` yaml
matrix:
  include:
    - rvm: '2.1'
      sudo: required
      services: docker
      env: PUPPET_VERSION="3.8.4" BEAKER_set="centos-6-docker"
      script: bundle exec rspec spec/acceptance/class_spec.rb
      bundler_args: --without development
  allow_failures:
    - env: PUPPET_VERSION="3.8.4" BEAKER_set="centos-6-docker"
```

Add new Beaker SUT: 

``` yaml
HOSTS:
  centos-6-x64:
    platform: el-6-x86_64
    hypervisor : docker
    image: centos:6
    docker_preserve_image: true
    docker_cmd: '["/sbin/init"]'
    docker_image_commands:
      - 'yum install -y crontabs tar wget'
CONFIG:
  type: foss
  log_level: debug
```

bastelfreak:
:+1: I like the idea, why not, can't be bad to have a fallback for PCCI

rnelson0:
Should be at least 3.8.5 tho, .4 had some issues. Why not ~> 3.8.0 or ~>
3.0 to keep up?
>
>

-- 

Rob Nelson
rnelson0@gmail.com

daenney:
I agree with @rnelson0. Our normal matrix specifies `~> 3.0` and `~> 4.0` for Puppet versions. We should stick to the same here.

nibalizer:
I think the camptocamp people do it so :+1:

jyaworski:
Agree. We have acceptance tests already, but they're not being run. As in, stuff in spec/acceptance.

bastelfreak:
We're now the proud owner of https://github.com/voxpupuli/puppet-puppetserver (form camptocamp). They do a lot of cool acceptance testing with .travis.yml and .travis.sh. We should fiddle that out soon because I don't want to kill the tests with a modulesync, but also I want to modulesync.

daenney:
Most of what happens in `travis.sh` seems to be related to getting specific versions of Augeas in place and testing with that. For modules that use Augeas this makes sense, but not by a long shot for all modules.

bastelfreak:
[...]

PR #241, opened 2016-10-13
Add support for additional parameters

dhoppe:
This is an requirement to restore the additional build steps (Beaker) of the Corosync module https://github.com/voxpupuli/puppet-corosync/pull/380.


dhoppe:
This will fix #80.

PR #244, opened 2016-10-19
require latest strings release

bastelfreak:


config_defaults.yml
```diff
@@ -42,7 +42,7 @@ Gemfile:
       - gem: voxpupuli-release
         git: https://github.com/voxpupuli/voxpupuli-release-gem.git
       - gem: puppet-strings
-        version: '0.4.0'
+        version: '~> 0.99.0'
```
bbriggs: Wouldn't you just leave the version key off to get the latest? Or is the aim to keep us at the latest below 1.0.0?

igalic:
we have  no real use of this? do projects have to declare their own in .travis.yml?

bastelfreak:
@igalic its for the fuuuuuuuuuuuuuuuuture. We're planning to use it in a few modules, for example nginx. Nothing is needed in the modules .travis.yml/.syc.yml for this to work.

igalic:
:sparkles: magic :sparkles:

PR #247, opened 2016-10-20
Add support for Coveralls

dhoppe:
This will allow us to use more badges at the REAMDE.md. In this case it will show use the percentage of the code coverage for RSpec tests.

badges

For example: https://coveralls.io/github/voxpupuli/puppet-collectd

We already use this for some of our repositories: https://coveralls.io/github/voxpupuli/puppet-lint-classes_and_types_beginning_with_digits-check


bastelfreak:
tested at https://travis-ci.org/voxpupuli/puppet-dhcp

PR #257, opened 2016-10-28
Enable simplecov-console coverage output

### LINKED ISSUE #256: Use simplecov-console
We've now got coveralls setup, but it'd be nice if coverage information was output to the console when running tests locally. https://github.com/chetan/simplecov-console

alexjfisher:
Fixes #256

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -58,6 +58,9 @@ Gemfile:
       - gem: coveralls
         ruby-operator: '>='
         ruby-version: '2.0.0'
+      - gem: simplecov-console
+        ruby-operator: '>='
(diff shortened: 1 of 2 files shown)


config_defaults.yml
```diff
@@ -58,6 +58,9 @@ Gemfile:
       - gem: coveralls
         ruby-operator: '>='
         ruby-version: '2.0.0'
+      - gem: simplecov-console
```
bastelfreak: I guess we don't need the gem on travis? possible to move it to the development group? Probably not because of 'SimpleCov::Formatter::Console'?

PR #258, opened 2016-11-01
add r10k

bastelfreak:


managed_modules.yml
```diff
@@ -46,6 +46,7 @@
 - puppet-puppetserver
 - puppet-rhsm
 - puppet-rundeck
+- puppet-r10k
```
dhoppe: This module does not exist in the voxpupuli namespace.

bastelfreak: we are in the process of migrating it

PR #260, opened 2016-11-03
drop testing on ruby225

bastelfreak:
I don't think that this is really useful. Puppet AIO ships ruby21 for
the agent, and the latest Ruby, which is used in all the modern
distributions, is ruby231.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -9,8 +9,6 @@
     env: PUPPET_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test
   - rvm: 2.1.9
     env: PUPPET_VERSION="~> 4.0" CHECK=test
-  - rvm: 2.2.5
-    env: PUPPET_VERSION="~> 4.0" CHECK=test
   - rvm: 2.3.1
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi, I would like to see some feedback here before we merge/decline this. CC: @voxpupuli/project-maintainers

domcleal:
The only common OSes I know of using 2.2.x are FreeBSD (using 2.2.5 I believe, @mmoll may be able to confirm), and Fedora 23, which was superseded by 24 and 25 Beta anyway.

bastelfreak:
Yep, 
FreeBSD11 still ships Puppet 4.6.2 which uses Ruby 2.2.5_1,1 :sadface:

mmoll:
Shipping a newer version of Puppet also wouldn't change the Ruby used. 2.2.5p319 is indeed the current default on FreeBSD.

jyaworski:
I disagree, here. https://www.ruby-lang.org/en/downloads/ is the current list of supported rubies. We should be trying to stick to that where possible, given legacy OS constraints and supported Puppet version constraints.

PR #262, opened 2016-11-05
Update list of supported Ruby versions

dhoppe:


bastelfreak:
I don't think that we [should use it](https://www.ruby-lang.org/en/news/2016/04/01/ruby-2-1-10-released/):
```
... This release is not intended for production use...

**Comment by dhoppe** [comment]
OK :(

**Comment by domcleal** [comment]
This would be a good place to use it I think, it's a test matrix and may ensure that the module + stack work with the two digit version number. It isn't being deployed to "production" here, it's under test.

**Comment by rnelson0** [comment]
There's not going to be a `2.1.11` so doesn't seem to hold much value; it would just be another travis worker occupied on every build.

**Comment by domcleal** [comment]
> it would just be another travis worker occupied on every build.

It's replacing 2.1.9, so no extra load.

**Comment by rnelson0** [comment]
Ah. Guess I'm ambivalent then. Is there a module(/branch where we added it so we know it works in at least one case?

**Comment by dhoppe** [comment]
@rnelson0 I tested 2.1.10 with all my modules and did not have any trouble.

**Comment by jyaworski** [comment]
Can we just say 2.1? RVM has some magic that resolves that to ~>2.1 IIRC.

**Comment by dhoppe** [comment]
@jyaworski Do you know if that also works for 1.9, 2.2 and 2.3?

**Comment by jyaworski** [comment]
@dhoppe it should.

**Comment by dhoppe** [comment]
@jyaworski I think this is not going to work, because Travis CI uses this [Docker image](https://quay.io/repository/travisci/travis-ruby) which includes the following Ruby versions:

```
travis@104e56fd8c69:~$ rvm alias list | grep -v jruby | sort
2.0 => ruby-2.0.0-p598
2.1 => ruby-2.1.5
2.2 => ruby-2.2.0
default => ruby-1.9.3-p551
``` 

It only installs newer versions from the [Ruby version cache](http://rubies.travis-ci.org) if they are defined at the matrix including minor versions like `2.1.10`.

I try to find out if there is a GitHub repository which contains the `Dockerfile`. In this case I could create an PR to update the list of Rubies and their aliases. But this still means that Travis CI has to maintain this file to get the latest Ruby versions.

PR #266, opened 2016-11-16
Revert "Fix SimpleCove filter"

alexjfisher:
This reverts commit 152e345be5669b598cbb3d26d8ddbffd88d4a7b4. https://github.com/voxpupuli/modulesync_config/pull/263

Since this is a revert and might be controversial, I don't want it merged unless there first is a consensus.
I think @bastelfreak thinks we should probably revert, but I'd like to see some other people comment too.


rnelson0:
I concur with this change.

PR #269, opened 2016-11-20
bump to latest ruby versions

bastelfreak:
the ruby team released new versions, they are already available on travis:
http://rubies.travis-ci.org/


dhoppe:
Did you see the discussion at #262?

bastelfreak:
I did. I tried to only declare major + minor version but that didn't work out, rvm resolved it to older versions.

PR #272, opened 2016-11-24
Improve simplecov

alexjfisher:
Don't run simplecov at all if the module has no `lib` directory.
This should fix a github coveralls integration issue.  Coveralls seems
to get upset and never reports back to travis if we send it reports
about 0 files.

Include all ruby files under `lib` in the coverage report, even those
with 0% coverage.

### DIFF
--- a/moduleroot/spec/spec_helper.rb
+++ b/moduleroot/spec/spec_helper.rb
@@ -2,7 +2,7 @@ require 'puppetlabs_spec_helper/module_spec_helper'
 require 'rspec-puppet-facts'
 include RspecPuppetFacts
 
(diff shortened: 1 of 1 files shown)


alexjfisher:
Tested on puppet-zabbix (a module containing types/providers) and puppet-jira (one with no lib dir).

rnelson0:
@alexjfisher you might want to submit this patch to puppet-module-skeleton as well, everyone seems to be struggling with the right way to use simplecov.

alexjfisher:
@rnelson0 Lets give it a few days first, in case any more tweaking is necessary.

PR #275, opened 2016-11-30
Exclude `vendor` path from simplecov

### LINKED ISSUE #274: Make reporting coverage optional
I don't really care about having coverage reports when I'm just trying to get a test working. It might be nice to be able to disable coverage by setting an environment variable like `NO_COVERAGE_REPORT` to true or something like that.

rnelson0:
Fixes #274


bastelfreak:
could you add .vendor as well? A lot of people us that.

PR #276, opened 2016-11-30
Allow for additional cops that a module may require

rnelson0:


rnelson0:
Specifically, puppet-r10k needs to set `Rspec/MultipleExpectations` to `Max: 2`

rnelson0:
Should I update version to 0.16.4 or 0.17.0?

PR #281, opened 2016-12-07
Remove selinux_config_mode from default facts

alexjfisher:
This fact doesn't always exist on systems and having it as a default
fact is masking strict variable issues.

### DIFF
--- a/moduleroot/spec/default_facts.yml
+++ b/moduleroot/spec/default_facts.yml
@@ -3,4 +3,3 @@ concat_basedir: "/tmp"
 ipaddress: "172.16.254.254"
 is_pe: false
 macaddress: "AA:AA:AA:AA:AA:AA"
-selinux_config_mode: "disabled"


alexjfisher:
Need a bit of feedback on this before merging.
It *will* cause builds to fail, and modules will need fixing.

All the same, I think it's right to remove `selinux_config_mode`.  Even though it's a core fact, it's not set if the `selinux` fact is false.
See https://github.com/voxpupuli/puppet-zabbix/pull/326

bastelfreak:
:+1: from my side, @voxpupuli/collaborators what do you think?

jyaworski:
@alexjfisher can you explain how it's masking issues? This was added precisely to solve some of those issues by setting facts that don't matter.

alexjfisher:
@jyaworski I was getting `Puppet Unknown variable: '::selinux_config_mode'` warnings from puppet-zabbix in my puppetserver logs.  Spec tests didn't fail when run with `STRICT_VARIABLES=yes` because the fact is assumed to always be available.  In reality, it isn't.  I think we test with strict vars on because puppet 5 will enforce this?

PR #288, opened 2016-12-22
add safe_yaml ~> 1.0.4

wyardley:
Let me know if this is the right place within the file
See:
https://tickets.puppetlabs.com/browse/PUP-3796
But basically, this seems to avoid a problem with rake where PUPPET_VERSION="~> 3.0" is set, but Ruby is >= 2.2.x

```
% bundle exec rake 
rake aborted!
NameError: uninitialized constant Syck
/usr/local/lib/ruby/gems/2.3.0/gems/puppet-3.8.7/lib/puppet/vendor/safe_yaml/lib/safe_yaml/syck_node_monkeypatch.rb:42:in `<top (required)>'
/usr/local/lib/ruby/gems/2.3.0/gems/puppet-3.8.7/lib/puppet/vendor/safe_yaml/lib/safe_yaml.rb:197:in `require'
/usr/local/lib/ruby/gems/2.3.0/gems/puppet-3.8.7/lib/puppet/vendor/safe_yaml/lib/safe_yaml.rb:197:in `<module:YAML>'
[...]
%  ruby --version
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]
```

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -61,6 +61,8 @@ Gemfile:
       - gem: simplecov-console
         ruby-operator: '>='
         ruby-version: '2.0.0'
+      - gem: safe_yaml
+        version: '~> 1.0.4'
     ':development':
       - gem: travis
       - gem: travis-lint


rnelson0:
Can you link to where this is showing up in Travis? Syck means a Ruby prior to 2.3 and we shouldn't be testing puppet3 with that.

wyardley:
@rnelson0 I don't know of it causing problems in Travis, but the error came up when trying to test a problem when PUPPET_VERSION was 3 and STRICT_VARIABLES was set.

Yes, using the exact Ruby version of the failing test would be ideal, but setting up rvm or other things for multiple Ruby versions seemed like overkill, and adding this solved my problem.

rnelson0:
The issue is that puppet 3 isn't being tested against Ruby 2.3 (https://github.com/puppetlabs/puppet/blob/3.x/.travis.yml) so while this may address your issue, it may also cause issues with certain modules. We did just announce plans to start up puppet3 branches in modules and this fix could be applied there, rather than in a global modulesync location that will be deprecated in about a week.

PR #289, opened 2016-12-23
build on master and tags

bastelfreak:
the filter matches all branches and tags, the description 'branches' is a bit misleading. The regex matches our tag schema and was successfully tested for the virtualbox release.
https://github.com/voxpupuli/puppet-virtualbox/pull/41


rnelson0:
This could build on branches with weird names but I think that's worth the minuscule risk.

PR #294, opened 2017-01-04
Bump Rubocop gems to specific versions

james-powis:
The pinning basically was pinned to 1.0.0 > Ver >= 0.39.0 on rubocop and 2.0.0 > Ver >= 1.6.0 on rubocop-rspec which is a bit reckless, As such I have pinned to latest minor allowing for patch, this mimics the rest of the gem dependencies in principle.


bastelfreak:
thanks!

PR #303, opened 2017-01-12
move main jobs to ruby240

bastelfreak:
the json people published json185. This is a new 1.X version that works
with ruby240. Ruby240 brings several speed improvements so jobs should
be finished faster. This was tested at
https://github.com/voxpupuli/puppet-splunk/pull/83


sacres:
This works for acceptance checks as well, provided: `PUPPET_INSTALL_TYPE=agent` is *_explicitly_* set. Otherwise: 8< ... 

> undefined method `install_pe_on'
...

Thanks!

PR #305, opened 2017-01-13
add pmtignore to exclude strings-docs

bastelfreak:
We need to discuss if we want this or not. In the past we had issues with builds that were too big so we might need to exclude the docs from releases.


rnelson0:
Docs don't do any good on my puppet master.

dhollinger:
Considering docs can be generated with a rake task, I don't see why we need them pulled from the Forge

vinzent:
Including the reason why it is excluded in the commit message would be nice. So if anybody in 6 months looks at it can see it right away from the git log.

http://chris.beams.io/posts/git-commit/#seven-rules

bastelfreak:
good point, updated the commit message.

PR #306, opened 2017-01-17
Rubocop updates

vinzent:
* fix in spec_helper for new rubocop test - use safe_load
* fix renamed rubocop checks


vinzent:
same as tested in puppet-selinux PR 182 https://github.com/voxpupuli/puppet-selinux/pull/182

PR #310, opened 2017-01-20
Use `YAML.safe_load`, re-enable Rubocop check

lamawithonel:
This commit replaces calls to `YAML.load` in `spec/spec_helper.rb` with
calls to `YAML.safe_load` and re-enables the related Rubocop check in
the modulesync_config base directory.  Prior to this, because the cop
was not disabled in the module root, modules synced to this repo would
fail Rubocop checks while this modulesync_config repo would pass.  A
change in code was prefered here, because the check is a security check
rather than style, lint, or metrics check, and the code change has no
effect on the outcome of the spec helper.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,7 +14,3 @@ Metrics/LineLength:
 # dealbreaker:
 Style/TrailingCommaInArguments:
   Enabled: false
-
-# Disable warnings on moduleroot/spec/spec_helper.rb and other files
-Security/YAMLLoad:
-  Enabled: false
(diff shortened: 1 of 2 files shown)


rnelson0:
This change breaks development with Ruby 2.0.0, which VP has decided to support because EL7 includes it. 

Additionally, `safe_load` suppresses exceptions on malformed YAML input. If we were accepting unsanitized input as on a web page, this would probably be a good idea, but since this is parsing our facts yaml sources, we actually do want to receive exceptions if malformed content is found in them.

lamawithonel:
I think disabling security checks per usage would be the best route, but if it needs to be disabled for platform support then that decision will have to wait. Closing.

PR #313, opened 2017-01-31
Add CentOS 6 and 7 beaker nodesets

vinzent:
These use vagrant boxes provided by the distro itself.

Puppetlabs boxes are out of date whereas the distro box is
maintained.

The main difference to the puppetlabs boxes is IMHO that
SELinux is enforcing (which is the default when installing
the OS). The puppetlabs boxes had SELinux disabled.


vinzent:
I'm using these nodesets on voxpupuli/puppet-selinux

PR #314, opened 2017-02-02
Add Amazon Linux EC2 example beaker nodeset

vinzent:
Beaker supports EC2 as hypervisor and Amazon provides some free
resources. Registration and credentials required.

Other distros than Amazon Linux could be added/configured.


vinzent:
i'll followup up with more nodesets.

PR #316, opened 2017-02-03
Update Fedora 25 beaker nodeset

vinzent:
Set platform to correct version as puppet-agent 1.9 officially
supports Fedora 25 now.


vinzent:
puppet-selinux acceptance tests ran succesfully against this updated nodeset.

PR #319, opened 2017-02-08
bump puppetlabs_spec_helper + add needed dependency

bastelfreak:
I tested this locally with puppet-zabbix. Works great so far, spec tests run now in parallel. However we should hold of with merging, puppetlabs_spec_helper seems to be broken in certain situations. https://github.com/puppetlabs/puppetlabs_spec_helper/pull/177

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -17,7 +17,8 @@ Gemfile:
   required:
     ':test':
       - gem: puppetlabs_spec_helper
-        version: '~> 1.2.2'
+        version: '~> 2.0.1'
+      - gem: parallel_tests
       - gem: rspec-puppet
         version: '~> 2.5'
       - gem: rspec-puppet-facts


rnelson0:
Have you reviewed how the output looks? See https://github.com/voxpupuli/puppet-collectd/issues/633 where I ran some tests against collectd. It's fine when everything is green, but when there are errors, it's ... not pretty.

bastelfreak:
In my opinion this is okay. It is harder to read, yes, but the speed gain is huge. We can still run the tests without parallel if it gets to ugly for debugging.

bastelfreak:
puppetlabs_spec_helper got released so in theorie we could merge.

config_defaults.yml
```diff
@@ -17,7 +17,8 @@ Gemfile:
   required:
     ':test':
       - gem: puppetlabs_spec_helper
-        version: '~> 1.2.2'
+        version: '~> 2.0.1'
+      - gem: parallel_tests
```
igalic: i don't understand why this gem isn't just a transitive dependency that we don't have to care for, because `puppetlabs_spec_helper` depends on it.

igalic: yes. https://github.com/puppetlabs/puppetlabs_spec_helper/blob/355162fb6afe19ee40b1c0dd9a169fb134032e96/README.md#parallel-tests

rnelson0:
Ah, an earlier draft of the feature made the test target always run in parallel, which was my concern. Glad it's optional.

PR #324, opened 2017-03-09
Mask tty's in docker acceptance nodesets

ekohl:
Since beaker containers are privileged these tty's can interfere with
the system tty's. This can be an unpleasant surprise for those running
on Linux.

### DIFF
--- a/moduleroot/spec/acceptance/nodesets/docker/centos-7.yml
+++ b/moduleroot/spec/acceptance/nodesets/docker/centos-7.yml
@@ -11,6 +11,7 @@ HOSTS:
     docker_cmd: '["/usr/sbin/init"]'
     docker_image_commands:
       - 'yum install -y crontabs initscripts iproute openssl sysvinit-tools tar wget which ss'
+      - 'systemctl mask getty@tty1.service'
 CONFIG:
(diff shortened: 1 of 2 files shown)


bastelfreak:
Is the tty required if we want to debug the container and enter it after a puppet run?

ekohl:
I'm not sure if docker attach actually uses a tty (I'd guess a pts), but /sbin/init is non-interactive anyway. Beaker uses SSH server to connect I used that instead. A small script (that assumes you have one running container):
```
#!/bin/sh
container=$(docker ps --quiet)
ip=$(docker inspect $container | jq '.[0]["NetworkSettings"]["IPAddress"]' -r)
ssh -o UserKnownHostsFile=/dev/null root@$ip
```

The password is root.

bastelfreak:
Ah correct, we've got ssh. Should we document the ssh script somewhere in the README.md or in the [CONTRIBUTING.md](https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.github/CONTRIBUTING.md) we place in each repository?

bastelfreak:
@ekohl could you please add the script later to one of the readmes?

PR #335, opened 2017-04-08
Provide a default overcommit config file

vinzent:
Hooks are only enabled if you take action.

To enable the hooks run:

```
bundle install
bundle exec overcommit --install
bundle exec overcommit --sign
```

(it will manage the .git/hooks directory):

Examples howto skip a test for a commit or push:

```
SKIP=RuboCop git commit
SKIP=PuppetLint git commit
SKIP=RakeTask git push
```

Don't invoke overcommit at all:

```
OVERCOMMIT_DISABLE=1 git commit
```

Read more about overcommit: https://github.com/brigade/overcommit

Things to discuss:

* what to run on PrePush (see hook list on github)
* what to run on PrePreCommit (see hook list on github)
* there are also few that are enabled by default (see github overcommit README.md
  - the ones with * prefix in the hook list)
* do we need to exclude some directories by default  (consider: the non-rake
  tasks are only run if a changed file matches an include rule)?
* i thought about making it configureable with .sync.yml but then decided to do
  it when its really required (maybe only few will use this overcommit thing
  because it needs manual action to enable it)

### DIFF
--- a/moduleroot/.overcommit.yml
+++ b/moduleroot/.overcommit.yml
@@ -1,0 +1,63 @@
+# Managed by https://github.com/voxpupuli/modulesync_configs
+#
+# Hooks are only enabled if you take action.
+#
+# To enable the hooks run:
+#
+# ```
+# bundle exec overcommit --install
+# # ensure .overcommit.yml does not harm to you and then
+# bundle exec overcommit --sign
+# ```
+#
+# (it will manage the .git/hooks directory):
+#
+# Examples howto skip a test for a commit or push:
+#
+# ```
+# SKIP=RuboCop git commit
+# SKIP=PuppetLint git commit
+# SKIP=RakeTask git push
+# ```
+#
+# Don't invoke overcommit at all:
+#
+# ```
+# OVERCOMMIT_DISABLE=1 git commit
+# ```
+#
+# Read more about overcommit: https://github.com/brigade/overcommit
+#
+# To manage this config yourself in your module add
+#
+# ```
+# .overcommit.yml:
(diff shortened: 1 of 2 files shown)


vinzent:
I've run `bundle exec msync update --namespace vinzent --message "Test modulesync" --filter puppet-selinux --branch modulesync_test`to test-apply the modulesync and it prevented me to push a change to puppet-selinux because of new puppet-lint errors.

moduleroot/.overcommit.yml
```diff
@@ -0,0 +1,63 @@
...
+#   unmanaged: true
+# ```
+#
+# to your modules .sync.yml config
+---
+PreCommit:
+  RuboCop:
+    enabled: true
+    description: 'Runs rubocop on modified files only'
+    command: ['bundle', 'exec', 'rubocop']
```
bastelfreak: should we use the rake task here? Or as an alternative:
bundle exec rubocop -c .rubocop.yml -D -S

vinzent: RakeTask rubocop will run on every commit.  RuboCop task will only run if there is a ruby file involved with the commit.

as commit needs to be fast i've choosen the builtin tasks instead of the file unaware rake-task.

vinzent: the command is maybe missleading. overcommit will not just run plain rubocop.

vinzent: https://github.com/brigade/overcommit/blob/master/config/default.yml#L504 is what it adds as params.

PR #336, opened 2017-04-15
Disable rubocop for Gemfile/Rakefile

bastelfreak:
we generate both files with modulesync. changing the layout with rubocop
will bring us a cyclic dependency. We need to fix our template
generation before we run rubocop on both files.


vinzent:
puppet-nginx: auto-correction 76 of 79 offenses
puppet-selinux auto-correction 64 of 72 offenses

PR #338, opened 2017-04-16
Ignore Guardfile when doing rubocop checks

juniorsysadmin:
```Style/FileName: The name of this source file (Guardfile) should use snake_case```
Guardfiles are really a personal preference thing and shouldn't be synced or checked

### DIFF
--- a/moduleroot/.gitignore
+++ b/moduleroot/.gitignore
@@ -17,6 +17,7 @@ Puppetfile.lock
 *.iml
 .*.sw?
 .yardoc/
+Guardfile
 <% if ! @configs['paths'].nil? -%>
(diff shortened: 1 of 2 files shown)


bastelfreak:
We ship Guardfiles in multiple repos. Should we put them in the .gitignore or should we manage it via modulesync?

juniorsysadmin:
.gitignore . Definitely don't manage via modulesync.

bastelfreak:
Can you update the moduleroot/.gitignore to  also list the Guardfile?

juniorsysadmin:
> Can you update the moduleroot/.gitignore to also list the Guardfile?

Done

PR #339, opened 2017-04-16
add locales to the ubuntu16.04 docker image

bastelfreak:
the package got dropped upstream from the default container, so we need
to reinstall it.


bastelfreak:
reference: https://travis-ci.org/voxpupuli/puppet-collectd/jobs/222576519#L1366

PR #342, opened 2017-04-25
Rename all templates to `.erb`

DavidS:
This is a recommended change after https://github.com/voxpupuli/modulesync/pull/114, and will help with reducing the amount of grief editors and other tools throw up with trying to interpret ERB tags.


domcleal:
This requires an unreleased version of modulesync, it might be safest to revert as even the new version will be backwards-compatible for a while. I'll try and prep a new release though.

DavidS:
We could also bump the modulesync dependency in the Gemfile to a git
reference for a while.

PR #343, opened 2017-04-28
Ignore naming of Guardfile

domcleal:
Guardfile:1:1: C: Style/FileName: The name of this source file (Guardfile)
    should use snake_case.  (https://github.com/bbatsov/ruby-style-guide#snake-case-files)


domcleal:
Already excluded from all cops in edaefb9.

PR #346, opened 2017-05-10
Disable Rubocop RSpec/BeforeAfterAll check

### LINKED ISSUE #345: Disable RSpec/BeforeAfterAll
Disable the RSpec/BeforeAfterAll check as it interferes with the `before(:context)` blocks that some modules' acceptance tests use to clean the test environment before running the next context.

`before(:context)` is an alias to `before(:all)` in rspec, hence why rubocop sees it

dhollinger:
fixes #345

### DIFF
--- a/moduleroot/.rubocop.yml.erb
+++ b/moduleroot/.rubocop.yml.erb
@@ -508,6 +508,10 @@ Style/ClosingParenthesisIndentation:
 
 # RSpec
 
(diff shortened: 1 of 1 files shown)


domcleal:
Could you use `Exclude` to only disable this for `spec/acceptance/` rather than globally? I think it's still a useful and easier rule to follow in unit tests.

PR #350, opened 2017-05-23
add the parallel_tests gem

bastelfreak:
this allows us to run spec tests in parallel. We already enabled this in
our configs in the past, but we didn't include this gem. It was a
dependency of puppetlabs_spec_helper. This got removed so we need to
install it explicitly. Changes tested at:
https://github.com/voxpupuli/puppet-zabbix/blob/master/Gemfile#L39
https://travis-ci.org/voxpupuli/puppet-zabbix/builds/235141105

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -74,6 +74,7 @@ Gemfile:
       - gem: github_changelog_generator
         ruby-operator: '>='
         ruby-version: '2.2.2'
+      - gem: parallel_tests
     ':development':
       - gem: travis
       - gem: travis-lint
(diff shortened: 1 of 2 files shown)


domcleal:
Could you also change the rspec output format to dots in that case, in `moduleroot/.rspec.erb` or `.rspec_parallel.erb` if you want to only change the parallel mode?

It should make the output much easier to read than the documentation style when there's a failure, otherwise the output becomes interleaved between the parallel processes and it's very hard to find the error.

domcleal:
Thanks @bastelfreak.

PR #353, opened 2017-05-28
(WIP) Create a github release

vinzent:
Additionlly to the puppet forge push also create a release
on github.

* needs an encrypted github access token 
* don't know what it writes as text or as title

### DIFF
--- a/moduleroot/.travis.yml.erb
+++ b/moduleroot/.travis.yml.erb
@@ -89,6 +89,7 @@ deploy:
   - provider: releases
     api_key: 
       secure: "GITHUB OAUTH TOKEN"
+    file_glob: true
     file: "pkg/*.tar.gz"
     skip_cleanup: true
     on:
--- a/moduleroot/.travis.yml.erb
+++ b/moduleroot/.travis.yml.erb
@@ -76,13 +76,21 @@ branches:
 notifications:
   email: false
 deploy:
-  provider: puppetforge
-  user: puppet
-  password:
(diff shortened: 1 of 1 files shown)


moduleroot/.travis.yml.erb
```diff
@@ -76,13 +76,22 @@ branches:
...
+      secure: "<%= @configs['secure'] -%>"
+    on:
+      tags: true
+      # all_branches is required to use tags
+      all_branches: true
+      # Only publish the build marked with "DEPLOY_TO_FORGE"
+      condition: "$DEPLOY_TO_FORGE = yes"
+  - provider: releases
+    api_key: 
+      secure: "GITHUB OAUTH TOKEN"
```
bastelfreak: I'm not sure who can generate this, probably only people with admin terms? We should move the generation of the token to plumbing and read it here from the .sync.yml (as we do it with the string in line 82)

rnelson0: I think since this would be a org oauth token, it would be the same for all modules.

vinzent: is the travis encryption of the key by repo or by org?

ghoneycutt:
@vinzent I don't get the point in this as when you submit a tag to github, then github creates a release. This code would only create releases when a tag exists. If a tag exists then wouldn't a github release also already exist?

ghoneycutt:
Closing due to lack of activity, please feel free to re-open.

PR #356, opened 2017-06-14
use correct https url in Gemfile, without .git

bastelfreak:


bastelfreak:
We got a bugreport from cy4n that .git suffix on an https url fails on his mac: https://gist.github.com/cy4n/d753a5d38d45346996e25a585cc49183

traylenator:
Sorry ... was a bit quick to merge there ...

PR #360, opened 2017-07-19
Update all the things for puppet5 + general cleanup

bastelfreak:


oranenj:
Not sure if we really want to drop tests for all older rubies, but if it makes our tests actually run, it's better than having the tests actually fail all the time...

bastelfreak:
If we notice that we need them we can still add them back (or just drop puppet4 support)

config_defaults.yml
```diff
@@ -10,32 +10,23 @@
     env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=@@SET@@ CHECK=beaker
     bundler_args: --without development
   includes:
-  - rvm: 2.1.9
```
alexjfisher: Puppet 4 AIO packages ship with this ruby, don't they?  I would have kept this.

PR #367, opened 2017-08-25
Add puppet-rabbitmq to managed modules

wyardley:


dhollinger:
looks like the travis failures are related to `puppet-system` rather than rabbitmq (puppet-system needs to be modernized)

PR #368, opened 2017-09-04
Don't set config.future_release for -rc versions

alexjfisher:
It can sometimes be nice to refresh the changelog with unreleased
changes.  When the rake task is run and the version is an `-rc` release
`config.future_release` will no longer be set.


rnelson0:
@alexjfisher something's wrong with this and puppet-system, but I'm not sure if it's with the module or changes to modulesync.

alexjfisher:
@rnelson0 I guess it was https://github.com/voxpupuli/puppet-system/pull/32 ?

PR #369, opened 2017-09-07
remove puppet-lint-absolute_classname-check from defaults

wyardley:
As discussed on IRC, maybe we want to consider dropping this check?

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -31,7 +31,6 @@ Gemfile:
         version: '~> 2.5'
       - gem: rspec-puppet-facts
       - gem: rspec-puppet-utils
-      - gem: puppet-lint-absolute_classname-check
       - gem: puppet-lint-leading_zero-check
       - gem: puppet-lint-trailing_comma-check
       - gem: puppet-lint-version_comparison-check


bastelfreak:
I"m not sure if we actually want this, I need to think about this.

vinzent:
@wyardley as GH readers don't know what was discussed on IRC it would IMHO make sense to include an abstract of it as commit+PR description.

https://chris.beams.io/posts/git-commit/

wyardley:
@vinzent You're right, sorry about that. Some folks, including @bastelfreak were suggesting using
`contain foo` rather than `contain ::foo` or something along those lines.

We had a discussion where several current Puppet employees said that there's no specific guideline requiring this in most cases.

The style guide does recommend using absolute scope in certain cases (such as when namespacing variables; https://docs.puppet.com/puppet/5.1/style_guide.html#namespacing-variables). As far as whether it's required to do:
`class '::foo'`
vs
`class foo`, I'm not sure if it's as clear-cut. Most of the examples in the style guide don't use it, e.g.,
https://docs.puppet.com/puppet/5.1/style_guide.html#separate-files
```
# init.pp
class apache { }
# ssl.pp
class apache::ssl { }
```

wyardley:
I'm personally totally fine either way. I'd been updating everything to use `::foo` for ages, because I had that lint module installed, and because most Vox modules require it. Personally, I think it looks a little nicer without it.

Not sure if it's possible to enforce it for variable names, but not for `class` / `contain` / `include` etc. It does seem to allow `Class['foo']`, which is how most examples in the style guide are as well.

bastelfreak:
I thought about this and I prefer to always use the top scope with leading ::

rnelson0:
As an FYI, scoping is only useful for variables. Class names are no longer EVER relative [starting with Puppet 4](https://tickets.puppetlabs.com/browse/PUP-121) (docs [old](https://docs.puppet.com/puppet/3/lang_namespaces.html#relative-name-lookup-and-incorrect-name-resolution) and [new](https://docs.puppet.com/puppet/4.1/lang_namespaces.html) for comparison), and thus `include|contain|require ::foo` is no longer necessary, even if when there is a `<module>::foo` class, unless for some reason we are supporting Puppet 3 again. I would vote to drop it overall.

wyardley:
Also some discussion in
https://github.com/voxpupuli/puppet-lint-absolute_classname-check/issues/3

rnelson0:
Re-opening this for even further discussion!

The current plugin works on classnames. Classnames should only be prepended with :: when using puppet 3, which VP no longer supports, thus I feel it should be removed from our configuration.

I am not aware of a plugin that works on variable names, but that is likely not something a linter can catch. You wouldn't want to prohibit users from referring to `$ntp::package_name` as `$package_name` inside the ntp class and forcing them to use `$::ntp::package_name` everywhere. The only reasonable thing I could see a linter doing is if it sees a global `$::package_name` reference inside a class as well as `$package_name`, it throws a warning on both because it's confusing - but it is not wrong or a violation of the style guide. I do not think it is something we run into often and would probably not be worth the false positives it would generate.

wyardley:
I didn't have as strong a feeling about this one way or another when I created it, but it seems that:

1. Vox is only supporting Puppet 4 on most / all modules now
1. Using ::foo vs foo in most cases is _less_ preferred with Puppet  4
1. The bugs that were the main reason for encouraging this practice don't exist anymore

So, I do think that a good case has been made that this check should be dropped in our modules

wyardley:
fwiw, puppet-lint already checks variables for top scope namespace I think
```
% puppet-lint xyz.pp 
WARNING: top-scope variable being used without an explicit namespace on line 117
WARNING: top-scope variable being used without an explicit namespace on line 121
```

PR #370, opened 2017-09-07
run acceptance tests on latest ruby version

bastelfreak:
We currently run all tests on 2.4.1. So this should be the default in
modulesync_config


alexjfisher:
From IRC
```
<afisher> Is that tested anywhere?
<bastelfreak> https://github.com/voxpupuli/puppet-zabbix/blob/master/.travis.yml
<bastelfreak> all acceptance tested I touched are running on 2.4.1
<bastelfreak> but that was before we set the ruby version via modulesync_config
```

PR #371, opened 2017-09-07
Python doesn't exist on the Ubuntu 16.04 base Docker node

wyardley:
Since Python 2 is installed on older images (and in the Vagrant images distributed by Puppet), what do we think about having this in our default nodeset? 

If not, I can work around it in https://github.com/voxpupuli/puppet-rabbitmq/pull/617


ekohl:
I'd actually 👎 this even though I like python. If modules assume python exists, you can run into problems. Only if we needed this ourselves for testing infra then I'd say yes but now it can mask actual problems.

PR #372, opened 2017-09-08
Revert "Python doesn't exist on the Ubuntu 16.04 base Docker node"

wyardley:
Reverts voxpupuli/modulesync_config#371
@ekohl @bastelfreak I'm Ok with reverting this.

It did seem odd to me that python2 isn't on the Docker image we use but is on the Vagrant image that Puppet ships for acceptance tests (and on the Ubuntu 14.04 images for both Docker and Vagrant), but I'm not sure what the normal  / typical case is for Ubuntu 16.04, and I agree that it might make sense to revert this.


ekohl:
@wyardley a general trend is to get tiny base images. Fedora is moving to a system-python package and with dnf being rewritten in C I wouldn't be surprised if a minimal RHEL 8 would have no python installed either.

Issue #177, opened 2016-06-06
Missing spec_helper_acceptance.rb ?


madAndroid:
On a fresh clone of this repo, following the instructions in the `README.md`, and attempting a sync, a get the following error:

```
~/r/v/modulesync_config git:master  ❯❯❯❯ bundle exec msync update -f jenkins --message "modulesync 0.7.0"
Syncing puppet-jenkins_job_builder
Overriding any local changes to repositories in ./modules
Switching to branch modulesync
bundler: failed to load command: msync (/srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/bin/msync)
Errno::ENOENT: No such file or directory @ rb_sysopen - ./moduleroot//spec/spec_helper_acceptance.rb
  /srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/gems/modulesync-0.6.1/lib/modulesync/renderer.rb:14:in `read'
  /srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/gems/modulesync-0.6.1/lib/modulesync/renderer.rb:14:in `build'
  /srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/gems/modulesync-0.6.1/lib/modulesync.rb:77:in `block (2 levels) in run'
  /srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/gems/modulesync-0.6.1/lib/modulesync.rb:68:in `each'
  /srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/gems/modulesync-0.6.1/lib/modulesync.rb:68:in `block in run'
  /srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/gems/modulesync-0.6.1/lib/modulesync.rb:57:in `each'
  /srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/gems/modulesync-0.6.1/lib/modulesync.rb:57:in `run'
  /srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/gems/modulesync-0.6.1/bin/msync:8:in `<top (required)>'
  /srv/data/repos/voxpupuli/modulesync_config/.vendor/ruby/2.1.0/bin/msync:23:in `load'
[...]

PR #379, opened 2017-09-16
Add beaker/module_install_helper

wyardley:
Any reason not to have beaker/module_install_helper by default? It really simplifies spec_helper_acceptance


ekohl:
Related I think we should revisit https://github.com/voxpupuli/modulesync_config/issues/177.

PR #382, opened 2017-09-17
Minor updates to contributing guidelines

wyardley:
Some proposed updates to the wording of the contributing guidelines.

Maybe we should also have a pointer to IRC and / or Slack?

If we can find a succinct example of how to squash / rebase, that would also be really helpful for new contributors.

### DIFF
--- a/moduleroot/.github/CONTRIBUTING.md.erb
+++ b/moduleroot/.github/CONTRIBUTING.md.erb
@@ -1,6 +1,6 @@
 This module has grown over time based on a range of contributions from
 people using it. If you follow these contributing guidelines your patch
-will likely make it into a release a little quicker.
+will likely make it into a release a little more quickly.
 
 ## Contributing
 
@@ -65,6 +65,10 @@ add tests if you're adding new functionality. If you've not used
 [rspec-puppet](http://rspec-puppet.com/) before then feel free to ask
 about how best to test your new feature.
 
+To run the linter, the syntax checker and the unit tests:
+
+    bundle exec rake test
+
 To run your all the unit tests
 
     bundle exec rake spec SPEC_OPTS='--format documentation'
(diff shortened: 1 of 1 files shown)


moduleroot/.github/CONTRIBUTING.md.erb
```diff
@@ -89,15 +89,20 @@ with:
...
-    BEAKER_set=centos-66-x64 bundle exec rake acceptances
+    bundle exec rake beaker:centos-7-x64
+    bundle exec rake beaker:ssh:centos-7-x64
 
-If you don't want to have to recreate the virtual machine every time you
-can use `BEAKER_DESTROY=no` and `BEAKER_PROVISION=no`. On the first run you will
-at least need `BEAKER_PROVISION` set to yes (the default). The Vagrantfile
-for the created virtual machines will be in `.vagrant/beaker_vagrant_fies`.
+If you don't want to have to recreate the virtual machine every time you can
+use `BEAKER_DESTROY=no` and `BEAKER_PROVISION=no`. On the first run you will at
```
ekohl: This should be ```BEAKER_destroy=no``` and ```BEAKER_provision=no```. On the next line as well.

moduleroot/.github/CONTRIBUTING.md.erb
```diff
@@ -1,6 +1,6 @@
 This module has grown over time based on a range of contributions from
 people using it. If you follow these contributing guidelines your patch
-will likely make it into a release a little quicker.
+will likely make it into a release a little more quickly.
```
ekohl: I'm not a native speaker but what's wrong with ```quicker```?

wyardley: I don't know how to explain, I think it just sounds too colloquial here? Could be wrong.

moduleroot/.github/CONTRIBUTING.md.erb
wyardley: You're right, I didn't add those originally but totally didn't notice the mistake either (funny since we were just talking about this).

PR #388, opened 2017-09-30
Add debian9 docker nodeset

ekohl:


bastelfreak:
Hi @ekohl, did you test that nodeset somewhere? I tried it on autofs:
https://travis-ci.org/voxpupuli/puppet-gluster/jobs/281706066

ekohl:
I tested this on theforeman-dns and there it worked. Not sure why it doesn't on gluster but I'm pretty sure it has to do with the PUPPET_VERSION var that's exported. Not sure why it doesn't fail on other distros but Puppetlabs has been very slow with supporting Debian Stretch. Wouldn't be surprised if that combo wasn't supported.

Issue #195, opened 2016-07-12
Move gems not needed for `test` rake task to new group.


alexjfisher:
For example...
At the moment, it's not possible to test against puppet versions < 3.7 as puppet-strings depends on this version.

Recently, puppet-blacksmith had rest-client pinned to 1.8 as 2.0 didn't support ruby 1.9.  If puppet-blacksmith had been in a group other than `test` this wouldn't have been so much of an issue.

juniorsysadmin:
+1
Adding a publishing gem group reduces gem dependencies for the testing set and reduces the chances that something breaks when just trying to run unit tests. And this helps to get Pull Requests merged quicker, especially when there is nothing wrong with the submitter's code.

bastelfreak:
we discussed this multiple times on IRC. Currently our release fails because of the github changelog generator and a changed dependency: https://travis-ci.org/voxpupuli/puppet-openvpn/jobs/284429243

We should create the release group, which contains at least:

* github_changelog_generator
* voxpupuli-release
* puppet-blacksmith
* puppet-strings?

PR #392, opened 2017-10-06
migrate release related gems in own group

bastelfreak:
This is an attempt to fix https://github.com/voxpupuli/modulesync_config/issues/195


config_defaults.yml
```diff
@@ -81,6 +69,19 @@ Gemfile:
       - gem: beaker-puppet_install_helper
       - gem: beaker-module_install_helper
         require: false
+    ':release':
+       - gem: github_changelog_generator
```
rnelson0: This is over indented it appears.

bastelfreak: good catch, fixed it.

PR #396, opened 2017-10-18
puppet-iis is deprecated; don't manage it

wyardley:
This should be taken out, I believe


ekohl:
I think we also need something in the repo description so you don't need to scroll down to see it.

PR #402, opened 2017-10-21
remove two more modules (bacula and report_hipchat) from management

wyardley:
Remove puppet-bacula and puppet-report_hipchat

### DIFF
--- a/managed_modules.yml
+++ b/managed_modules.yml
@@ -3,7 +3,6 @@
 - puppet-amanda
 - puppet-autofs
 - puppet-archive
-- puppet-bacula
 - puppet-boolean
 - puppet-cassandra
 - puppet-collectd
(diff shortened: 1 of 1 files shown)


managed_modules.yml
```diff
@@ -58,7 +57,6 @@
 - puppet-puppetserver
 - puppet-rabbitmq
 - puppet-r10k
-- puppet-report_hipchat
```
alexjfisher: https://github.com/voxpupuli/puppet-report_hipchat/pull/26 should get this going again.

managed_modules.yml
```diff
@@ -3,7 +3,6 @@
 - puppet-amanda
 - puppet-autofs
 - puppet-archive
-- puppet-bacula
```
alexjfisher: +1 I think module needed a complete rewrite

PR #421, opened 2017-11-17
Update Fedora 25, 26 and add Fedora 27 nodeset

vinzent:
... and the Fedora 26-beta nodeset is removed.


bastelfreak:
Rerun travis, not it is green...

PR #422, opened 2017-11-17
Add a task to print the supported nodesets

ekohl:


bastelfreak:
Thanks for this! Very smart change \o/

ekohl:
Ideally I'd hook this into modulesync so you use this instead of ```.sync.yml``` to generate ```.travis.yml``` but I have yet to look into that.

PR #425, opened 2017-11-17
Add more tooling

ekohl:


ekohl:
Updated to clean Ubuntu 12.04 too.

bastelfreak:
Should we clean fedora 24 and older as well?

ekohl:
Makes sense.

ekohl:
Updated to prune Fedora and Red Hat too.

bin/clean-metadata
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+import os
+import json
+
+# Note range is exclusive so the last number is not in the list
+UNSUPPORTED = {
+    'CentOS': [str(i) for i in range(3, 6)],
+    'Debian': [str(i) for i in range(3, 6)],
+    'Fedora': [str(i) for i in range(3, 25)],
```
bastelfreak: Fedora 25 still has support. 24 and older is dead.

ekohl: Note the comment above: range is exclusive so 25 is not included.

bastelfreak: ah ups. but then we should increase debian to 7 (or even 8)?

ekohl: You're right. 8 it is.

ekohl:
Expanded to prune all unsupported Ubuntu releases:
```
Reading modules/puppet-logrotate/metadata.json
Removing CentOS-5
Removing Debian-6
Removing Debian-7
Removing Ubuntu-12.04
Removing Ubuntu-10.04
Writing modules/puppet-logrotate/metadata.json
```

bastelfreak:
Thanks for the work @ekohl!

PR #429, opened 2017-11-18
add lldpd

bastelfreak:


bastelfreak:
lol. so the server offers https, but redirects to http. apt doesn't like that and fails. WTF... interesting that this works in yum.

PR #430, opened 2017-11-18
always mock the systemd fact

bastelfreak:
The fact comes from camptocamp-systemd. We mock it in so many places
that it should be centralized. This got tested on puppet-lldpd from @ekohl and @bastelfreak

### DIFF
--- a/moduleroot/spec/spec_helper.rb.erb
+++ b/moduleroot/spec/spec_helper.rb.erb
@@ -48,6 +48,20 @@ RSpec.configure do |c|
   <%- end -%>
 end
 
+on_supported_os.each do |_os, facts|
+  systemd_fact = case facts[:os]['family']
+                 when 'Archlinux'
+                   true
+                 when 'RedHat'
+                   facts[:os]['release']['major'].to_i >= 7 ? true : false
+                 when 'Debian'
+                   facts[:os]['release']['major'].to_i >= 8 ? true : false
+                 else
+                   false
+                 end
+  add_custom_fact :systemd, systemd_fact
+end
+
 <%- [@configs['spec_overrides']].flatten.compact.each do |line| -%>
 <%= line %>
 <%- end -%>


TraGicCode:
First things first.  It seems like some of this should be tucked away in some sort of helper function to keep the spec_helper.rb clean and small.  I'm by far no ruby expert but the best thing i came up with is the following.

# spec/spec_helper.rb
```
# automatically load any facts we want stubbed across all modules
Dir['./spec/support/facts/*.rb'].sort.each { |f| require f }


on_supported_os.each do |_os, facts|
   add_custom_fact :systemd, get_systemd_fact_value(facts)
 end
```


 # spec/support/facts/systemd_fact.rb
```
def get_systemd_fact_value(facts)
    systemd_fact = case facts[:os]['family']
    when 'Archlinux'
      true
    when 'RedHat'
      facts[:os]['release']['major'].to_i >= 7 ? true : false
    when 'Debian'
      facts[:os]['release']['major'].to_i >= 8 ? true : false
    else
      false
    end
end
```


We could even take this a step further and have some sort of conditional check that says if the metadata.json contains the camptocamp-systemd module then lets auto stub this.  Just some input.  Now saying we should do this :D

TraGicCode:
This folder might make more sense for the fact stubbing function

spec/support/facts/camptocamp-systemd/systemd.rb

bastelfreak:
@ekohl what do you think about this? IMO we should fine a solution here that works for theforeman and us.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -48,6 +48,20 @@ RSpec.configure do |c|
   <%- end -%>
 end
 
+on_supported_os.each do |_os, facts|
+  systemd_fact = case facts[:os]['family']
+                 when 'Archlinux'
+                   true
+                 when 'RedHat'
+                   facts[:os]['release']['major'].to_i >= 7 ? true : false
```
ekohl: The ternary operating is redundant since the comparison is already a boolean.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -48,6 +48,20 @@ RSpec.configure do |c|
...
+                 when 'Debian'
+                   facts[:os]['release']['major'].to_i >= 8 ? true : false
```
ekohl: Same here

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -48,6 +48,20 @@ RSpec.configure do |c|
...
+                 else
+                   false
+                 end
+  add_custom_fact :systemd, systemd_fact
```
ekohl: I don't think you can call ```add_custom_fact``` multiple times with the same fact. I think it overrides it so you end up with the last value in the loop.

ekohl:
I like the idea of auto stubbing if a dependency is present. Not sure how feasible it is right now, but really nice. Maybe it's something for the library where every module can provide a ```facterdb.rb``` in the module root. That would automatically solve it when you add a dependency and not require special care in spec helpers. It does make me wonder if we're not replicating facter then but in cases where you call a binary it does make sense to mock it.

ekohl:
Inspired by the above I created https://github.com/voxpupuli/modulesync_config/pull/431 which I think solves a lot of the problems.

ekohl:
Closing due to inactivity.

PR #431, opened 2017-11-19
[WIP] Improve custom fact handling

ekohl:


moduleroot/spec/facts/puppetlabs/stdlib.rb
```diff
@@ -0,0 +1,18 @@
+add_custom_fact :puppet_environmentpath, '/etc/puppetlabs/code/environments'
+add_custom_fact :puppet_vardir, '/opt/puppetlabs/puppet/cache'
+add_custom_fact :root_home, '/root'
+add_custom_fact :service_provider, ->(os, facts) do
```
bastelfreak: what does `->` do and how is this operator called?

bastelfreak:
Thanks for the PR. We should test it on a module to see that it actually works (and call it acceptance-testing for msync)

moduleroot/spec/facts/puppetlabs/stdlib.rb
ekohl: I based it on https://github.com/mcanevet/rspec-puppet-facts/blob/fe21de12108fbab0123bdc4db2365a29ea62f171/spec/rspec_puppet_facts_spec.rb#L576-L580 so I guess it's called a proc :)

moduleroot/spec/facts/puppetlabs/stdlib.rb
```diff
@@ -0,0 +1,18 @@
...
+  case facts[:osfamily]
+  when 'RedHat'
+    facts[:operatingsystemrelease].to_i >= 7 ? 'systemd' : 'sysv'
```
ekohl: I was also offline so I don't know if it's actually sysv or something else.

ekohl:
What I used to verify it:

```
require 'spec_helper'

on_supported_os.each do |os, facts|
  context "on #{os}" do
    it { expect(facts["service_provider"]).not_to be_nil }
  end
end
```

moduleroot/spec/facts/puppetlabs/stdlib.rb
```diff
@@ -0,0 +1,18 @@
+add_custom_fact :puppet_environmentpath, '/etc/puppetlabs/code/environments'
+add_custom_fact :puppet_vardir, '/opt/puppetlabs/puppet/cache'
+add_custom_fact :root_home, '/root'
```
juniorsysadmin: Do we need to account for FreeBSD/OpenBSD tests?

ekohl: Maybe. I just went through what some modules set as standards but they probably disregarded those as well. Also no idea what it does on Windows.

ghoneycutt:
Seems we already have this functionality with https://github.com/mcanevet/rspec-puppet-facts 

https://github.com/voxpupuli/modulesync_config/blob/master/config_defaults.yml#L39

ekohl:
No, we don't. Those are the built in Puppet facts. This is intended to mock custom facts from modules. I'm still not sure if I like this format. Ideally the mocked fact could live inside the modules themselves so we don't need to modulesync it everywhere.

ghoneycutt:
Closing due to inactivity, please feel free to re-open.

PR #440, opened 2018-01-03
add pdk gem to development group

bastelfreak:


ekohl:
Should we drop redundant dependencies from our gem file then?

bastelfreak:
Sounds like a good idea, then pdk should not be in the development group, maybe in no group at all.

bastelfreak:
@ekohl I checked https://github.com/puppetlabs/pdk/blob/master/pdk.gemspec#L24-L36, pdk ships less deps than I thought. I think we can just add it to our development group and thats it?

ghoneycutt:
Why would we need pdk to be installed as a gem?

ghoneycutt:
Closing due to inactivity, please feel free to re-open.

PR #443, opened 2018-01-04
Add script to count unreleased commits + cleanup

bastelfreak:


bin/create-pull-requests
```diff
@@ -7,7 +7,7 @@ fi
 
 version=$(git describe)
 for module in modules/* ; do
-  pushd $module
+  pushd "$module" > /dev/null || exit
```
ekohl: How about ```#!/bin/bash -e``` then?

bin/count-unreleased-commits
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+for module in modules/* ; do
+  pushd "$module"  > /dev/null || exit
```
ekohl: Rather than using pushd you could also use ```git --git-dir``` I think.

bastelfreak: fixed

bin/create-pull-requests
bastelfreak: done

ekohl: You can remove the || exit since that's what ```-e``` does.

bin/count-unreleased-commits
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+for module in modules/* ; do
+  #pushd "$module"  > /dev/null || exit
```
ekohl: Can you remove the comments?

bin/create-pull-requests
bastelfreak: done

bin/count-unreleased-commits
bastelfreak: done

PR #446, opened 2018-01-28
bump ruby 2.4.2->2.4.3

bastelfreak:


vinzent:
@bastelfreak why?

bastelfreak:
In general we always try to test against the latest ruby versions. We did those changes many time in the past as well (valid versions for travis are listed at http://rubies.travis-ci.org/ ).

ekohl:
I do realize now that Puppet bundles a Ruby version. It might be good to match it exactly.

vinzent:
@ekohl @bastelfreak IHMO puppet-agent AIO included/supported ruby and latest ruby would be good to test.

unfortunatly the puppetserver ruby (jruby) is still on 1.9 level IMHO which would be important for serverside code (ruby parser functions mainly I think -  as long as you use `puppet generate types`)

bastelfreak:
ah that vendor foo....
Let's see what we get here: https://tickets.puppetlabs.com/browse/PUP-8402

PR #447, opened 2018-01-28
bump puppetlabs_spec_helper 2.5.0->2.6.0

bastelfreak:


bastelfreak:
to be honest, I can't remember why we started that

bastelfreak:
pinning to ~> 2.6 should be fine as well?

PR #459, opened 2018-03-30
add fail2ban module

bastelfreak:


managed_modules.yml
```diff
@@ -20,6 +20,7 @@
 - puppet-dropbear
 - puppet-etherpad
 - puppet-extlib
+- puppet-fail2ban
```
ekohl: Shouldn't this be below facette if you sort it?

bastelfreak:
@ekohl updated it

PR #461, opened 2018-04-10
Pin mocha version to ~> 1.4.0 for now

wyardley:
It seems like 1.5.0 creates some breaking changes, so we should probably pin the default to 1.4.x until we get a handle on this? Let me know if the version specification should be a range instead.


wyardley:
See also https://github.com/voxpupuli/puppet-jenkins/pull/865

PR #462, opened 2018-04-19
Minor portability fixes

smortex:
These commits fix a bunch of portability problems in the shell scripts contained in this repo.

Feel free to squash the commits: I tried to explain each point but such a detail level is maybe not necessary 😉

### DIFF
--- a/bin/create-pull-requests
+++ b/bin/create-pull-requests
@@ -1,4 +1,6 @@
-#!/bin/bash -e
+#!/bin/sh
+
+set -e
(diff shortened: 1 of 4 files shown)


bin/create-pull-requests
```diff
@@ -1,13 +1,14 @@
-#!/bin/bash -e
```
ekohl: https://github.com/voxpupuli/modulesync/pull/147 would make this script redundant

PR #464, opened 2018-04-28
run tests on latest ruby versions

bastelfreak:


bastelfreak:
This got tested on https://github.com/voxpupuli/puppet-lldpd/pull/41

PR #466, opened 2018-04-28
Ensure the puppet version requirement is correct

ekohl:


bastelfreak:
tested this locally, works.

PR #453, opened 2018-02-23
Use beaker-hostgenerator for docker acceptance tests

ekohl:
Depends on https://github.com/puppetlabs/beaker-hostgenerator/pull/109 and will require changes to modules with docker tests. Also allows removal of the files in the actual containers.

It'd be nice if we could also push the EC2 and vagrant boxes into beaker-hostgenerator.

* [ ] https://github.com/voxpupuli/puppet-autofs/pull/118
* [ ] https://github.com/voxpupuli/puppet-bird/pull/11
* [ ] https://github.com/voxpupuli/puppet-collectd/pull/791
* [ ] https://github.com/voxpupuli/puppet-corosync/pull/441
* [ ] https://github.com/voxpupuli/puppet-gluster/pull/166
* [ ] https://github.com/voxpupuli/puppet-grafana/pull/108
* [ ] https://github.com/voxpupuli/puppet-jira/pull/246
* [ ] https://github.com/voxpupuli/puppet-lldpd/pull/44
* [ ] https://github.com/voxpupuli/puppet-nginx/pull/1216
* [ ] https://github.com/voxpupuli/puppet-nodejs/pull/367
* [ ] https://github.com/voxpupuli/puppet-nscd/pull/37
* [ ] https://github.com/voxpupuli/puppet-php/pull/452
* [ ] https://github.com/voxpupuli/puppet-prometheus/pull/190
* [ ] https://github.com/voxpupuli/puppet-puppetboard/pull/212
* [ ] https://github.com/voxpupuli/puppet-rabbitmq/pull/702
* [ ] https://github.com/voxpupuli/puppet-report_hipchat/pull/36
* [ ] https://github.com/voxpupuli/puppet-rsyslog/pull/79
* [ ] https://github.com/voxpupuli/puppet-rundeck/pull/376
* [ ] https://github.com/voxpupuli/puppet-splunk/pull/181
* [ ] https://github.com/voxpupuli/puppet-squid/pull/96
* [ ] https://github.com/voxpupuli/puppet-yum/pull/98
* [ ] https://github.com/voxpupuli/puppet-zabbix/pull/498

PR #467, opened 2018-04-29
add ubuntu1804 docker nodeset

bastelfreak:


ekohl:
Can we please finish https://github.com/voxpupuli/modulesync_config/pull/453 instead an avoid the whole noise in all modules? This does mean we need to get a PR into beaker-hostgenerator for 1804 [here](https://github.com/puppetlabs/beaker-hostgenerator/blob/master/lib/beaker-hostgenerator/data.rb#L1046-L1054)

bastelfreak:
thanks for reminding @ekohl !

PR #475, opened 2018-05-20
Add instructions for msync with octokit PRs

bastelfreak:


bastelfreak:
@ekohl this was a copy and paste issue. I'm not sure if we should remove the hub setup, maybe we need it in the future again.

PR #479, opened 2018-07-07
unpin puppet-strings

bastelfreak:
there is no reason to pin this. Months ago there were some issues with a
0.9 release wich we didn't want to pull in. Now 2.1.0 got released and
works without issues.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -91,7 +91,7 @@ Gemfile:
       - gem: voxpupuli-release
         git: https://github.com/voxpupuli/voxpupuli-release-gem
       - gem: puppet-strings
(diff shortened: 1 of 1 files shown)


config_defaults.yml
```diff
@@ -91,7 +91,6 @@ Gemfile:
       - gem: voxpupuli-release
         git: https://github.com/voxpupuli/voxpupuli-release-gem
       - gem: puppet-strings
-        version: '~> 1.0'
```
ekohl: Should we change it to `>=` perhaps?

bastelfreak: done

PR #490, opened 2018-08-10
Fix beaker dependencies

bastelfreak:


Gemfile
```diff
@@ -19,6 +19,6 @@ group :development do
   gem 'travis-lint'
 end
 
-gem 'modulesync'
+gem 'modulesync', git: 'https://github.com/voxpupuli/modulesync.git'
```
ekohl: Why do we need modulesync from git? Do we need a specific patch? I'd prefer to work to get a release out if needed.

bastelfreak: the code to open up PRs is still unreleased. Yes, a new release is the preferred solution

PR #496, opened 2018-09-05
run acceptance tests on puppet6-nightly

bastelfreak:
This was tested on:
* https://github.com/voxpupuli/puppet-zabbix/pull/533
* https://github.com/voxpupuli/puppet-rabbitmq/pull/723

### DIFF
--- a/moduleroot/.travis.yml.erb
+++ b/moduleroot/.travis.yml.erb
@@ -46,11 +46,13 @@ matrix:
 <% end -%>
 <% if @configs['docker_sets'] -%>
 <%   @configs['docker_sets'].each do |set| -%>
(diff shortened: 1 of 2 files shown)


moduleroot/.travis.yml.erb
```diff
@@ -26,6 +26,8 @@ before_install:
 script:
   - 'bundle exec rake $CHECK'
 matrix:
+  allow_failures:
+    - env: BEAKER_PUPPET_COLLECTION=puppet6-nightly
```
ekohl: I'd prefer to start without this so we can actually see it fail. I know I wont open Travis to see if those failed and I think it'll still be silently ignored

PR #500, opened 2018-10-03
Revert "add puppet-puppet_webhook"

bastelfreak:
This reverts commit 64ba7b2fa278c3889dddbe8d8eb0cc96991b6b5f.


bastelfreak:
We use pdk for this module. it doesn't work with msync.

PR #501, opened 2018-10-04
puppet6 is released, use it with acceptance tests

Dan33l:
Now puppet6 is realsed it is better to not use nightly as default.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -13,6 +13,12 @@
     sudo: required
     dist: trusty
     services: docker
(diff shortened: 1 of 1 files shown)


bastelfreak:
Should we instead add another travis slot for this? @dhollinger recommended this. We could test on puppet6 + puppet6-nightlies to catch issues early.

Dan33l:
Yes i agree. I probably misunderstood a resource limit about CI slots.
Done

ekohl:
IMHO the nightly testing could be good if we have cron runs of modules but AFAIK we don't do so now. That'd mean we have to get a PR in and just get lucky to see something break.

bastelfreak:
@ekohl I would like to have that, but who has time for that to configure it :(. I think nevertheless we should test the nightlies and afterwards try to enable the crons in travis

PR #505, opened 2018-10-06
Various improvements to the spec_helper

ekohl:
Otherwise it won't be used

### DIFF
--- a/moduleroot/spec/spec_helper.rb.erb
+++ b/moduleroot/spec/spec_helper.rb.erb
@@ -1,3 +1,6 @@
+# This file is managed via modulesync
+# https://github.com/voxpupuli/modulesync
+# https://github.com/voxpupuli/modulesync_config
 <%- if @configs['mock_with'] -%>
 RSpec.configure do |c|
   c.mock_with <%= @configs['mock_with'] %>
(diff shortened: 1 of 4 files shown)


moduleroot/spec/spec_helper.rb.erb
```diff
@@ -1,3 +1,8 @@
+<%- if @configs['mock_with'] -%>
+RSpec.configure do |c|
+  c.mock_with <%= @configs['mock_with'] %>
```
bastelfreak: this block is now duplicate. We probably can remove the other one. This need to be tested within one module

ekohl: I just verified this on puppet-mongodb (in my local branch) that the later one is indeed redundant and it must be done here at the top.

ekohl:
Updated to include various updates. See the individual commits for details.

bastelfreak:
can you please run this against one module and link the PR to this one?

ekohl:
https://github.com/voxpupuli/puppet-zabbix/pull/549

ekohl:
https://github.com/voxpupuli/puppet-nginx/pull/1264 as well

PR #508, opened 2018-10-08
Add puppet-vault_lookup

pcarlisle:


pcarlisle:
Is this even desirable? I started with pdk and can stick with that if it's better.

PR #510, opened 2018-10-21
Add Dockerfile for local tests

jkroepke:
Running rspec tests without setup a local environment. 

Just run `docker build .` on the root from the module to see the test results.

### DIFF
--- a/moduleroot/Dockerfile
+++ b/moduleroot/Dockerfile
@@ -1,0 +1,21 @@
+FROM ruby:2.5.1
+
+WORKDIR /opt/puppet
+
+# https://github.com/puppetlabs/puppet/blob/06ad255754a38f22fb3a22c7c4f1e2ce453d01cb/lib/puppet/provider/service/runit.rb#L39
+RUN mkdir -p /etc/sv
+
+ARG PUPPET_VERSION="~> 6.0"
+ARG PARALLEL_TEST_PROCESSORS=4
+
+# Cache gems
+COPY Gemfile .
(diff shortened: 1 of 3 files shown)


moduleroot/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+ARG PARALLEL_TEST_PROCESSORS=4
+
+# Cache gems
+COPY Gemfile .
+RUN bundle install --without system_tests development release --path=${BUNDLE_PATH:-vendor/bundle}
+
+COPY . .
+
+RUN bundle install
+RUN bundle exec rake rubocop
```
ekohl: `bundle exec release_checks` covers all tests, including linting which you've now skipped I think.

bastelfreak: updated it

ekohl:
We should add this to the `.pmtignore` so it's not shipped to the forge but otherwise I'd be OK with this.

bastelfreak:
@ekohl updated it.

bastelfreak:
I gave it a try and it worked. If any user has high experiences in docker he might come up with enhancements. Otherwise I think it's fine.

dhoppe:
@ekohl The number of files is not important. It is much more about the number of layers which have been reduced by using ```bundle exec release_checks```.

ekohl:
@dhoppe I was talking about the number of files we modulesync. I want to keep that number as low as possible.

Recently you might have seen me remove a lot of them by making sure we can generate nodesets on the fly (with beaker-hostgenerator). This means we only need to update one gem rather than n modules. It also means the entire ecosystem can benefit.

PR #511, opened 2018-10-24
pin parallel_tests for ruby 2.1.0

bastelfreak:
This was tested in https://github.com/voxpupuli/puppet-nginx/pull/1270


bastelfreak:
This fails because of https://github.com/voxpupuli/puppet-mongodb/pull/514

bastelfreak:
okay it maybe fails because of something differen :(

alexjfisher:
For completeness, the 'something different' was fixed in https://github.com/voxpupuli/puppet-mongodb/pull/515

PR #518, opened 2018-11-30
Use BEAKER_HYPERVISOR env var

ekohl:
Beaker 4.2.0 allows specifying the hypervisor explicitly via an env var.  This means we can simplify the setfile and allow complex definitions via .sync.yml


ekohl:
https://github.com/voxpupuli/puppet-mongodb/pull/519 is green now

PR #520, opened 2018-12-02
rename dockerfile

bastelfreak:


dhoppe:
@bastelfreak Why have you renamed this file?  It does not contain any ERB syntax.

traylenator:
I also questioned this but every other file is erb extension already so decided it was consistent.

bastelfreak:
all files in modulesync_config/moduleroot/ have to be .erb files since some time. No matter if they are an actual config or a static file.

PR #522, opened 2018-12-16
Refactor default fact loading

ekohl:
This gets rids of the default facts because our modules do not rely on them.

More importantly it switches to using add_custom_fact so we don't break the built in facts from rspec-puppet-facts.

Tests running in https://github.com/voxpupuli/puppet-r10k/pull/476. Likely to break some modules, but it's a better end result.


bastelfreak:
@ekohl can you please rebase?

PR #528, opened 2018-12-27
add dependencies for get_all_diffs script; refactor GitHub API handling

bastelfreak:


bin/get_all_the_diffs
```diff
@@ -26,10 +27,11 @@ DEBUG = false
...
 puts "get data from the github API" if DEBUG
-github_data1 = JSON.load(open('https://api.github.com/orgs/voxpupuli/repos?per_page=100&page=1'))
-github_data2 = JSON.load(open('https://api.github.com/orgs/voxpupuli/repos?per_page=100&page=2'))
-repos = github_data1.map{|repo| repo['name']} + github_data2.map{|repo| repo['name']}
-repos = repos.select{|repo| repo =~ /^puppet-/}.reject{|repo| repo =~ /^puppet-lint/}
+client = Octokit::Client.new
+client.auto_paginate = true
+gh_repos = client.repos('voxpupuli')
+repos = []
+gh_repos.map{|repo| repos << repo[:name] if repo[:name] =~ /^puppet-(?!lint)/}
```
ekohl: Can also be written as `repos = gh_repos.map{|repo| repo[:name]}.filter{|name| name =~ /^puppet-(?!lint)/}}`. If not, I'd use `each` since map is intended to return something.

bastelfreak: Updated it. I hat .map and .filter in the first version, but I wanted to reduce it to a single iteration.

PR #529, opened 2018-12-29
switch acceptance tests from trusty to xenial

bastelfreak:


ekohl:
At the top level we set the dist to Xenial so can't we leave the dist out altogether?

PR #530, opened 2019-01-01
Add ability to generate REFERENCE.md and test it can be generated

ghoneycutt:
See https://puppet.com/docs/puppet/6.1/modules_documentation.html#concept-3848


moduleroot/Rakefile.erb
```diff
@@ -41,6 +41,11 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
   t.pattern = 'spec/acceptance'
 end
 
+desc 'Generate REFERENCE.md'
+task :reference do
+  sh 'puppet strings generate --format markdown'
```
bastelfreak: should we call this via `bundle exec`? I don't think that everybody has puppet installed in `$PATH` at their workstations. At least I don't.

ekohl: I assume that you'd execute `bundle exec rake reference` then and it'd work that way.

ekohl: Can you make this configurable? In https://github.com/voxpupuli/puppet-extlib#development it's slightly different.

ghoneycutt: what needs to be configurable?

ghoneycutt:
@bastelfreak anything left here?

config_defaults.yml
```diff
@@ -116,6 +116,7 @@ Rakefile:
   default_enabled_rake_targets:
   - 'metadata_lint'
   - 'release_checks'
+  - 'reference'
```
ekohl: Not all modules generate meaningful references. I'm not sure we should add this by default just yet.

moduleroot/Rakefile.erb
ekohl: > Ah good catch. We also used this in the past on other modules:
> 
> ```
> bundle exec rake strings:generate\[',,,,false,true']
> ```

TIL: there's `require 'puppet-strings/tasks'` which gives you this task. Probably a good addition to puppetlabs_spec_helper to attempt to include this but silently skip it if unavailable.

I'm also thinking about contributing `strings:generate` to https://github.com/puppetlabs/puppet-strings instead of solving it in our modulesync.

config_defaults.yml
ghoneycutt: It does not hurt to generate it even if they don't have great output. It does hurt the project by not conforming to the Forge standards and creating the REFERENCE.md output. I think we should generate them, even if they are not 100% awesome and then work toward improving them.

moduleroot/Rakefile.erb
ghoneycutt: I think these ideas are great, but not directly impacted by this pull request. I don't see a compelling reason to not merge because one module does something different, when it should be generating the docs like every other module.

config_defaults.yml
ekohl: I think we could use a way to generate some overview of all our modules to get some insight. https://github.com/voxpupuli/puppet-lint-param-docs can check but https://github.com/voxpupuli/puppet-lint-param-docs/pull/11 would be an improvement. Also some tools to convert from the old docs format would greatly improve things.

Another issue I have is that I'm not a fan of including generated files in git because it's so much noise in git log and a great way to get merge conflicts.

moduleroot/Rakefile.erb
ekohl: I still want this to be configurable. We know puppet-extlib already needs customized options and it's trivial to add a variable that would make this possible.

config_defaults.yml
ghoneycutt: The rake task just shows that the docs can be generated. It is not meant to be run with each pull request. The actual generation of docs that is committed to git would be done as part of the release process.

+1 to using https://github.com/voxpupuli/puppet-lint-param-docs though that's a separate issue as it forces compliance. 

This PR is a step toward that, without having to force compliance with that linting. If you have docs, you should get the generated file.

There is no need to convert old docs as the REFERENCE.md file would be new and does not conflict with README.md. If a README lists all the params, their values, etc, we should remove that in favor of using REFERENCE.md, though the two are not tightly coupled.

First we should implement the ability to generate REFERENCE.md, then we should update README's and document code, then enforce that with the linting to ensure the quality continues.

moduleroot/Rakefile.erb
ghoneycutt: I don't get why it needs customizable options.

config_defaults.yml
ekohl: > The rake task just shows that the docs can be generated. It is not meant to be run with each pull request. The actual generation of docs that is committed to git would be done as part of the release process.

That's a misinterpretation. This config value means to run these checks when running `rake test`. IMHO generating the `REFERENCE.md` when testing isn't the right thing to do.

> There is no need to convert old docs as the REFERENCE.md file would be new and does not conflict with README.md. If a README lists all the params, their values, etc, we should remove that in favor of using REFERENCE.md, though the two are not tightly coupled.

I didn't mean that way of old docs (separate issue). A lot of our docs still use the old format (puppet-doc?) instead of yardoc. puppet-strings doesn't know how to deal with that and generates garbage. We can probably come up with an automatic converter to save a lot of effort.

> First we should implement the ability to generate REFERENCE.md, then we should update README's and document code, then enforce that with the linting to ensure the quality continues.

Right, but just adding the task to Rakefile.erb is sufficient for that. This config line is't needed.

moduleroot/Rakefile.erb
ekohl: To quote the earlier linked extlib README:

```
Reference documentation is generated using puppet-strings. To regenerate it, please run the rake task as follows.

 % bundle exec rake strings:generate\['lib/puppet/functions/extlib/*.rb,,,,false,true']
```

This is done to avoid generating docs for our deprecated functions.

ghoneycutt: Gotcha.. though if the functions still exist, they should have documentation.

ghoneycutt:
Not sure how to make it configurable. Suggest opening a new PR based on this and adding your changes there.

moduleroot/Rakefile.erb
ekohl: To be explicit, I was thinking about `sh 'puppet strings generate --format markdown<%= strings_options %>'` or something like that. Then extlib can add arguments in .sync.yml and we'll have a single command will just work in all modules even if the implementation is slightly different.



> I'm also thinking about contributing `strings:generate` to https://github.com/puppetlabs/puppet-strings instead of solving it in our modulesync.

https://github.com/puppetlabs/puppetlabs_spec_helper/pull/274 (added benefit that PDK users will also get this)

ekohl: https://github.com/puppetlabs/puppet-strings/pull/192 is what I actually meant. It doesn't solve the patterns issue though.

ekohl:
> Not sure how to make it configurable. Suggest opening a new PR based on this and adding your changes there.

See https://github.com/voxpupuli/modulesync_config/pull/543

ekohl:
Closing in favor of #543

Issue #151, opened 2016-05-08
Add rake task to detect trailing whitespace in markdown files


bastelfreak:
IMO it is best practice to avoid trailing whitespace. I'm not sure if any of our tasks already can handle this or if there are existing solutions. Any opinions?

bastelfreak:
maybe we also want to think a step further and offer something to automatically remove the whitespaces?

jyaworski:
@bastelfreak isn't this checked for by puppet-lint and rubocop both?

bastelfreak:
Our existing rake tasks seem to ignore it. I never get complaints when I run them local, just my vim does.

dhoppe:
Our current setup already does that. 

**Puppet lint**
```
dhoppe@macbook ~/Projects/puppet/puppet-kafka (git)-[master] % git diff
diff --git a/manifests/init.pp b/manifests/init.pp
index 33edd6a..e3389d5 100644
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -47,7 +47,7 @@
 #
 #
 class kafka (
-  $version        = $kafka::params::version,
+  $version        = $kafka::params::version,
   $scala_version  = $kafka::params::scala_version,
   $install_dir    = $kafka::params::install_dir,
   $mirror_url     = $kafka::params::mirror_url,
```

```
dhoppe@macbook ~/Projects/puppet/puppet-kafka (git)-[master] % bundle exec rake test
manifests/init.pp:50:trailing_whitespace:ERROR:trailing whitespace found
bundle exec rake test  4.78s user 0.27s system 99% cpu 5.077 total
```

**Rubocop**
```
dhoppe@macbook ~/Projects/puppet/puppet-kafka (git)-[master] % git diff
diff --git a/spec/classes/broker_spec.rb b/spec/classes/broker_spec.rb
index a1c4e7b..ef588ef 100644
--- a/spec/classes/broker_spec.rb
+++ b/spec/classes/broker_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 describe 'kafka::broker', type: :class do
   let :facts do
     {
-      osfamily: 'Debian',
+      osfamily: 'Debian',
       operatingsystem: 'Ubuntu',
       operatingsystemrelease: '14.04',
       lsbdistcodename: 'trusty',
```

```
dhoppe@macbook ~/Projects/puppet/puppet-kafka (git)-[master] % rubocop .
Inspecting 15 files
.....C.........

[...]

PR #532, opened 2019-01-02
(GH-151) Check for trailing whitespace in Markdown files

ghoneycutt:
This PR can easily be changed to add other files that should be checked
for trailing whitespace.

### DIFF
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -1,3 +1,4 @@
+# rubocop:disable Style/Not, Style/NegatedIf, Metrics/LineLength, Style/InverseMethods, Style/Next, Style/RegexpLiteral
 require 'puppetlabs_spec_helper/rake_tasks'
 
 # load optional tasks for releases
@@ -39,6 +40,24 @@ end
 desc 'Run acceptance tests'
 RSpec::Core::RakeTask.new(:acceptance) do |t|
   t.pattern = 'spec/acceptance'
+end
+
+desc 'Check for trailing whitespace'
+task :trailing_whitespace do
+  puts "\nChecking for trailing whitespace"
+  Dir.glob('**/*.md', File::FNM_DOTMATCH).sort.each do |f|
+    if not f =~ /^(modules\/|acceptance\/|\.?vendor\/|spec\/fixtures\/|pkg\/|REFERENCE.md)/
+      puts "  #{f}"
+      cnt = 0
+      File.foreach(f) do |line|
+        cnt += 1
+        if line =~ /\s\n$/
+          puts "  #{f} has trailing whitespace on line #{cnt}"
+          exit 1
+        end
+      end
+    end
+  end
 end
 
 <% targets = @configs['default_enabled_rake_targets']  - ( @configs['extra_disabled_rake_targets'] || [] ) -%>
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -116,6 +116,7 @@ Rakefile:
   default_enabled_rake_targets:
   - 'metadata_lint'
   - 'release_checks'
+  - 'trailing_whitespace'
 spec/default_facts.yml:
   delete: true
 spec/classes/coverage_spec.rb:


moduleroot/Rakefile.erb
```diff
@@ -41,6 +41,24 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
...
+desc 'Check for trailing whitespace'
+task :trailing_whitespace do
+  puts "\nChecking for trailing whitespace"
+  Dir.glob('**/*.md', File::FNM_DOTMATCH).sort.each do |f|
+    if not f =~ /^(modules\/|acceptance\/.*|\/vendor\/)/
+      puts "  #{f}"
+      cnt=0
+      File.foreach(f) do |line|
+        cnt += 1
+        if line =~ / $/
```
ekohl: I'd check for `\s` so you also catch tabs

ghoneycutt:
@bastelfreak now the commit is verified :)

moduleroot/Rakefile.erb
```diff
@@ -41,6 +41,24 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
   t.pattern = 'spec/acceptance'
 end
 
+desc 'Check for trailing whitespace'
+task :trailing_whitespace do
+  puts "\nChecking for trailing whitespace"
+  Dir.glob('**/*.md', File::FNM_DOTMATCH).sort.each do |f|
+    if not f =~ /^(modules\/|acceptance\/.*|\/vendor\/)/
```
bastelfreak: I testes this on the zabbix module:
```
$ bundle exec rake trailing_whitespace

Checking for trailing whitespace
  .github/CONTRIBUTING.md
  .github/ISSUE_TEMPLATE.md
  .github/PULL_REQUEST_TEMPLATE.md
  .vendor/ruby/2.5.0/bundler/gems/github-changelog-generator-5d7f18bc415e/CHANGELOG.md
  .vendor/ruby/2.5.0/bundler/gems/github-changelog-generator-5d7f18bc415e/CONTRIBUTING.md
  .vendor/ruby/2.5.0/bundler/gems/github-changelog-generator-5d7f18bc415e/CONTRIBUTING.md has trailing whitespace on line 40
```

Might it be a good idea to exclude every directory that is in our .gitignore file?

ghoneycutt: `.gitignore` uses a different matching system than ruby, so it would be non-trivial to set that up. I can work on adding exclusion patterns though.

ghoneycutt: Added `.vendor`

moduleroot/Rakefile.erb
```diff
@@ -41,6 +41,24 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
   t.pattern = 'spec/acceptance'
 end
 
+desc 'Check for trailing whitespace'
+task :trailing_whitespace do
+  puts "\nChecking for trailing whitespace"
+  Dir.glob('**/*.md', File::FNM_DOTMATCH).sort.each do |f|
+    if not f =~ /^(modules\/|acceptance\/|\/vendor\/|\.vendor\/|spec\/fixtures\/|pkg\/)/
```
bastelfreak: I need to correct myself. Is the `/` before `vendor` actually required? Because:

```
Checking for trailing whitespace
  .github/CONTRIBUTING.md
  .github/ISSUE_TEMPLATE.md
  .github/PULL_REQUEST_TEMPLATE.md
  CHANGELOG.md
  CONTRIBUTING.md
  README.md
  files/zapache/README.md
  vendor/bundle/ruby/2.4.0/bundler/gems/github-changelog-generator-5d7f18bc415e/CHANGELOG.md
  vendor/bundle/ruby/2.4.0/bundler/gems/github-changelog-generator-5d7f18bc415e/CONTRIBUTING.md
  vendor/bundle/ruby/2.4.0/bundler/gems/github-changelog-generator-5d7f18bc415e/CONTRIBUTING.md has trailing whitespace on line 40
The command "bundle exec rake $CHECK" exited with 1.
```
From https://travis-ci.org/voxpupuli/puppet-zabbix/jobs/475829901#L1482

bastelfreak: This works for me:

```
desc 'Check for trailing whitespace'
task :trailing_whitespace do
  puts "\nChecking for trailing whitespace"
  Dir.glob('**/*.md', File::FNM_DOTMATCH).sort.each do |f|
    next unless f !~ /^(modules\/|acceptance\/|vendor\/|\.vendor\/|spec\/fixtures\/|pkg\/)/
    puts "  #{f}"
    cnt = 0
    File.foreach(f) do |line|
      cnt += 1
      if line =~ %r{\s\n$}
        puts "  #{f} has trailing whitespace on line #{cnt}"
        exit 1
      end
    end
  end
end
```

Also rubocop did some reformatting

ghoneycutt: I don't agree with using `unless` as it makes logic harder to understand

`next unless !~` -- that's jibberish comparred to `if not matching`

ghoneycutt:
@bastelfreak fixed the `/vendor` issue and style around `cnt=0`.

ghoneycutt:
@ekohl we good? I made the changes you requested regarding updating the matching to catch tabs.

moduleroot/Rakefile.erb
```diff
@@ -41,6 +42,24 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
   t.pattern = 'spec/acceptance'
 end
 
+desc 'Check for trailing whitespace'
+task :trailing_whitespace do
+  puts "\nChecking for trailing whitespace"
+  Dir.glob('**/*.md', File::FNM_DOTMATCH).sort.each do |f|
+    if not f =~ %r{^(modules\/|acceptance\/|vendor\/|\.vendor\/|spec\/fixtures\/|pkg\/)}
+      puts "  #{f}"
```
ekohl: Does it need to be verbose by default?

moduleroot/Rakefile.erb
```diff
@@ -41,6 +42,24 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
   t.pattern = 'spec/acceptance'
 end
 
+desc 'Check for trailing whitespace'
+task :trailing_whitespace do
+  puts "\nChecking for trailing whitespace"
+  Dir.glob('**/*.md', File::FNM_DOTMATCH).sort.each do |f|
+    if not f =~ %r{^(modules\/|acceptance\/|vendor\/|\.vendor\/|spec\/fixtures\/|pkg\/)}
```
mmerfort: I don't think you need to escape the forward slashes, when you use %r{}

moduleroot/Rakefile.erb
ghoneycutt: Yes, the verbosity is useful.

bastelfreak: I think it is actually helpful. This helps us to detect if the condition in line 49 works or not.

moduleroot/Rakefile.erb
```diff
@@ -41,6 +42,24 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
   t.pattern = 'spec/acceptance'
 end
 
+desc 'Check for trailing whitespace'
+task :trailing_whitespace do
+  puts "\nChecking for trailing whitespace"
+  Dir.glob('**/*.md', File::FNM_DOTMATCH).sort.each do |f|
+    if not f =~ %r{^(modules/|acceptance/|vendor/|\.vendor/|spec/fixtures/|pkg/)}
```
ekohl: It's common in Ruby to use `unless` instead of `if not`.

You can also simplify the regex: `%r{^(modules|acceptance|\.?vendor|spec/fixtures|pkg)/}`.

Should we also ignore `REFERENCE.md` since that's a generated file?

moduleroot/Rakefile.erb
```diff
@@ -41,6 +42,24 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
...
+      puts "  #{f}"
+      cnt = 0
+      File.foreach(f) do |line|
```
ekohl: Alternative:
```ruby
File.foreach(f).each_with_index do |line, line_number|
```

moduleroot/Rakefile.erb
ghoneycutt: `unless` is not great in any language because it is not as clear as `if not`

I'll simplify the regex and move back to `//` instead of `%r{}` and will add `REFERENCE.md` which is a good catch.

ekohl: The alternative is `next if ....` or `if ... next end`

ghoneycutt: Or we can quit arguing about this pointless bit. I chose this small issue as a way to get to know this repo and give back to the VP community. The code works and the poster of the original issue, @bastelfreak, has already signed off on it. Please merge and then you can mangle the ruby as you see fit or create a new fork based off my work and co-author[1] this in a new PR.

I get that you want to ensure commits meet your standards though it seems this already meets the rest of VP's standards. This continual back and forth on a trivial contribution does not encourage community members to keep contributing.

[1] - https://help.github.com/articles/creating-a-commit-with-multiple-authors/

ekohl: > I get that you want to ensure commits meet your standards though it seems this already meets the rest of VP's standards.

I disagree with this. In all our repositories we follow rubocop to get a consistent coding style. You are disabling those checks and I'm not aware of precedents on this.

I am strict because this is synced to many repositories. It creates a lot of noise to change it so I strongly prefer to do it right the first time.

moduleroot/Rakefile.erb
```diff
@@ -41,6 +42,24 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
...
+desc 'Check for trailing whitespace'
+task :trailing_whitespace do
+  puts "\nChecking for trailing whitespace"
+  Dir.glob('**/*.md', File::FNM_DOTMATCH).sort.each do |f|
+    if not f =~ /^(modules\/|acceptance\/|\.?vendor\/|spec\/fixtures\/|pkg\/|REFERENCE.md)/
+      puts "  #{f}"
+      cnt = 0
+      File.foreach(f) do |line|
+        cnt += 1
+        if line =~ /\s\n$/
```
ekohl: Interesting. The `\n$` is redundant but $ also triggers on the end of the string. I suspect it'll fail on mac encoded files since they use `\r` but `/s$/` should catch those as well I think.

ekohl: It does look like `\s$` actually causes another issue to disregard this. We don't want mac encoded files anyway :)

ekohl:
https://github.com/voxpupuli/modulesync_config/pull/541 takes a different approach by implementing it as `check:trailing_whitespace` and enhancing `release_checks`. This means you can run it standalone, but is also executed by default.

PR #535, opened 2019-01-03
switch to latest ruby for puppet6 + support latest bundler release

bastelfreak:
This got tested at https://github.com/voxpupuli/puppet-zabbix/pull/587

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -3,17 +3,17 @@
   docker_sets:
   docker_defaults:
     # values will replace @@SET@@ with the docker_sets' value
-  - rvm: 2.5.1
+  - rvm: 2.5.3
     sudo: required
     services: docker
     env: PUPPET_INSTALL_TYPE=agent BEAKER_IS_PE=no BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_debug=true BEAKER_setfile=@@SET@@ BEAKER_HYPERVISOR=docker CHECK=beaker
     bundler_args: --without development release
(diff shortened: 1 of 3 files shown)


config_defaults.yml
```diff
@@ -22,10 +22,10 @@
...
     bundler_args: --without system_tests development release
-  - rvm: 2.5.1
+  - rvm: 2.5.3
     env: PUPPET_VERSION="~> 6.0" CHECK=test_with_coveralls
     bundler_args: --without system_tests development release
-  - rvm: 2.5.1
+  - rvm: 2.5.3
     env: PUPPET_VERSION="~> 6.0" CHECK=rubocop
     bundler_args: --without system_tests development release
   - rvm: 2.4.4
```
dhoppe: We should also update this version as long as the sync of the modules has not been done yet. The latest available stable version is 2.4.5.

config_defaults.yml
```diff
@@ -3,17 +3,17 @@
   docker_sets:
   docker_defaults:
     # values will replace @@SET@@ with the docker_sets' value
-  - rvm: 2.5.1
+  - rvm: 2.5.3
```
ekohl: If you always want the latest 2.5, why don't we just use `2.5`?

bastelfreak: The original idea was to use the same ruby version as in the AIO version. But I'm not sure how useful that actually is.

ekohl: In this case it doesn't matter because here we only launch a docker container

PR #536, opened 2019-01-05
bump rubocop target version 1.9->2.4

bastelfreak:
Our lowest supported Puppet version is 5.5.8. Puppet AIO shipped Ruby
2.4.4:
https://puppet.com/docs/puppet/5.5/about_agent.html
So we should tell rubocop to lint against this version.
Tested at https://github.com/voxpupuli/puppet-zabbix/pull/588


ekohl:
I'm not that up to date on Ruby, but I know that there were some actual syntax changes (way) in the past. If there was any old Ruby agent (even for another module) then it couldn't compile its catalog anymore because all ruby code was plugin synced. Will there be any of those problems here?

alexjfisher:
I've started a table of versions [here](https://github.com/voxpupuli/plumbing/issues/197).  I think it'll be quite useful when reviewing PRs like this one.  For instance...
Solaris *finally* shipped puppet 5, but is still on ruby 2.3.

baurmatt:
Should we get this merged? :) If I'm not mistaken, we currently require all our modules to support only Puppet 5 and greater which has already a dependency to Ruby 2.4 -> https://puppet.com/docs/puppet/5.5/system_requirements.html

alexjfisher:
@baurmatt The puppetserver ruby version is *not* 2.4 though.  Puppet server 5 defaults to jruby 1.7 (which is ruby 1.9 IIRC).  If we update rubocop, chances are when we 'autofix' failing cops, we'll create functions that no longer work.

bastelfreak:
In general I would like to get this merged. But it will also introduce a lot of required changes for every module and I currently don't have any time for this. Also there is still the issue with legacy puppetserver. I think that's EOL in May next year?

alexjfisher:
Puppet 5 is not EOL until Nov 2020

baurmatt:
> The puppetserver ruby version is not 2.4 though. Puppet server 5 defaults to jruby 1.7 (which is ruby 1.9 IIRC).

Interesting, I didn't know that. Thanks for explaining! :)

> But it will also introduce a lot of required changes for every module and I currently don't have any time for this.

Totally get this! :) Just stumble over this while looking for something in the rubocop file and was wondering about it.

> Puppet 5 is not EOL until Nov 2020

Then lets put this date in our calenders and put this issue on hold until then.

PR #542, opened 2019-01-11
Remove duplicate metadata_lint check

ekohl:
Since puppetlabs_spec_helper 1.1.0 this has been part of the validate task. Validate is part of release_checks.

It's also possible to fully remove this mechanism since no module appears to be using it. We can just standardize on release_checks instead of our custom task test but I chose the simple way here.


ekohl:
Btw, https://github.com/puppetlabs/puppet-strings/pull/192 is similar, but doesn't allow overriding the pattern via the config.

Issue #397, opened 2017-10-18
Create mechanism for auto-pushing generated docs


wyardley:
Lots of good discussion in #373

I'm going to close that PR, since it seems like things may be headed in a different direction; easy to bring back if we end up wanting to ignore the docs/ directory.

bastelfreak:
One idea:

I've a pre_release rake task in mind which bumps the module version in metadata.json to a provided version and afterwards generates the changelog. We could here also regenerate the docs.

Benefits:
* Simplifies release task
* Doc updates are up2date for each release
* Doc updates are reviewed before a release
* Clean git history

Cons:
* Docs aren't up2date after a regular pull request is merged. but that may not be needed

bastelfreak:
@voxpupuli/collaborators some feedback is welcome here :)

dhollinger:
I like this idea as it makes sure that the new docs are generated with the metadata bump and changelog generation.

seems to me like the most logical place to put it at this time

ekohl:
To confirm: when you release, the prepare_release:(patch|minor|major) task generates the docs + changelog and does the version bump. That's submitted in a PR. When you merge it, you create a tag and push it to Github which triggers Travis to push to the forge.

That would work for me.

Note that in the situation we don't bump the patch version after tagging but I'd be fine with that. You do need to ensure there's a prepare_release task that can skip the bumping but that's an implementation detail.

bastelfreak:
Our travis_relase task checks if the version in metadata.json has no -rc0 and that a matching entry is in the CHANGELOG.md. Afterwards it creates a tag (which gets released by travis). Then it bump the version metadata.json by one patch version and appends -rc0.

rnelson0:
Should this be part of travis_release, or added to blacksmith?

bastelfreak:
@rnelson0 I've no objections on this.

bastelfreak:
I currently hacked this in a modules Rakefile:

```ruby
desc 'prepare a new release'
task :prepare_release, [:version] do |t, args|
  # check for missing parameter
  # exit early, exit often
  (puts 'you need to provide a version like: rake prepare_releasep[1.0.0]'; exit) unless args[:version]
  version = args[:version]
  (puts 'format needs to be X.X.X'; exit) unless /\d+.\d+.\d+/.match(version)
  ENV['BLACKSMITH_FULL_VERSION'] = version
  Rake::Task['module:bump:full'].invoke
  Rake::Task['changelog'].invoke
[...]

PR #543, opened 2019-01-11
Add a reference task to generate REFERENCE.md

ekohl:
In puppet-extlib we can now set:

```yaml
Rakefile:
  puppet_strings_patterns: lib/puppet/functions/extlib/*.rb
```

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -39,7 +39,7 @@ Gemfile:
   required:
     ':test':
       - gem: puppetlabs_spec_helper
-        version: '>= 2.11.0'
+        version: '>= 2.14.0'
       - gem: rspec-puppet-facts
         version: '>= 1.8.0'
       - gem: rspec-puppet-utils
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -90,7 +90,7 @@ Gemfile:
       - gem: voxpupuli-release
         git: https://github.com/voxpupuli/voxpupuli-release-gem
       - gem: puppet-strings
-        version: '>= 1.0'
+        version: '>= 2.2'
 Rakefile:
   config.user: voxpupuli
   default_disabled_lint_checks:
@@ -101,6 +101,7 @@ Rakefile:
   - 'disable_single_quote_string_with_variables'
   default_enabled_rake_targets:
   - 'release_checks'
+  puppet_strings_patterns: []
 spec/default_facts.yml:
   delete: true
 spec/classes/coverage_spec.rb:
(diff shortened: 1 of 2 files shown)


ekohl:
I left it out here, but I think we can now modify our Travis release job as follows:

```
before_deploy: "bundle exec rake reference"
```

That should fix part of https://github.com/voxpupuli/modulesync_config/issues/397 by ensuring we always have an up to date REFERENCE.md on the forge for released modules.

ekohl:
Btw, https://github.com/puppetlabs/puppet-strings/pull/192 is similar, but doesn't allow overriding the pattern via the config.

moduleroot/Rakefile.erb
```diff
@@ -5,7 +5,16 @@ require 'puppetlabs_spec_helper/rake_tasks'
...
   require 'voxpupuli/release/rake_tasks'
+rescue LoadError
+end
+
+begin
   require 'puppet-strings/tasks'
+
+  desc 'Generate REFERENCE.md'
+  task :reference do
+    Rake::Task['strings:generate'].invoke('<%= @configs['puppet_strings_patterns'] %>', nil, nil, nil, 'false', 'true')
```
ghoneycutt: This should document what the fields are or link to the documentation as this is hard to grok.

ekohl: You're right. I initially attempted to pass in arguments as a named hash but it looks like the rake invoke task doesn't support this so I took this short cut.

ekohl:
Updated. This now requires the freshly released puppet-strings 2.2.0 which includes my patch with a `strings:generate:reference`. That makes our invocation a lot easier and self documenting.

ekohl:
https://github.com/voxpupuli/puppet-extlib/pull/131 is how puppet-extlib would make use of this.

config_defaults.yml
```diff
@@ -101,6 +101,7 @@ Rakefile:
   - 'disable_single_quote_string_with_variables'
   default_enabled_rake_targets:
   - 'release_checks'
+  puppet_strings_patterns: ''
```
alexjfisher: Should we make this an array instead?

ekohl: I like that idea. Will update

logicminds:
This should be

```
Rakefile:
  puppet_strings_patterns: lib/puppet/functions/**/*.rb functions/**/*.pp
```

ekohl:
Not sure I follow. Why is that?

logicminds:
Ahh nevermind that is the custom config for the extlib.

PR #548, opened 2019-02-01
cleanup travis.sh helper

bastelfreak:


ghoneycutt:
So simple we should just get rid of this script and put it back in the `.travis.yml` file. With it being so trivial, breaking this into another file adds complexity to the travis configuration that is not needed.

dhoppe:
@bastelfreak We should add the following to ```config_defaults.yml``` to completely remove the ```.travis/setup.sh```. 

```
.travis/setup.sh:
  delete: true
```

PR #555, opened 2019-03-06
Make parameters config.user / user configurable

dhoppe:
This should not be a static value to make sure that other projects can use msync.


dhoppe:
@ekohl Thank you for pointing that out. I have made the necessary changes.

PR #556, opened 2019-03-06
Make parameter user configurable

dhoppe:
This should not be a static value to make sure that other projects can use msync.


dhoppe:
@ekohl I will close this PR because the necessary changes are included in PR #555.

PR #557, opened 2019-03-08
Make notifications configurable

dhoppe:
This change allows other projects to configure notifications via e-mail or IRC according to their own needs. Otherwise we may receive their notifications in our IRC channel. 

The following entries in the file ```.sync.yml``` configure an e-mail address and deactivate the notifications via IRC.

```
.travis.yml:
  email: "notify@example.com"
  irc: false
```


ekohl:
Thanks!

PR #563, opened 2019-03-25
Add default mocking framework

dhoppe:
This will fix the deprecation warning:

```
puppetlabs_spec_helper: defaults `mock_with` to `:mocha`. See https://github.com/puppetlabs/puppetlabs_spec_helper#mock_with to choose a sensible value for you
```


bastelfreak:
I also think that this is a good approach. We can still configure it differently for specific modules if they require it.

PR #565, opened 2019-04-06
Clean Rakefile.erb

ekohl:
dacad9a Simplify the exclude_paths logic

   By always defining a default we can simplify our code and the defaults become a reference documentation.

623a46b Reuse the puppetlabs_spec_helper lint_fix task

   While it's a duplication, our task actually has a better self-documenting name.

702ad3e Simplify lint configuration

   This removes all the duplication that already happens in puppetlabs_spec_helper.

   It also renamed extra_disabled_lint_checks to a more saner puppet_lint_checks. This is safe because no module actually uses this. It's sanfer because disabled implies we're actually disabling them for you.

06be76b Hard alias test to release checks

   No module uses this and it only complicates our Rakefile.

586f9f1 Remove beaker_sets task

   Since we stopped storing beaker nodesets as files in our modules, this task no longer makes sense.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -93,12 +93,7 @@ Gemfile:
         version: '>= 2.2'
 Rakefile:
   config.user: voxpupuli
-  default_disabled_lint_checks:
-  - 'relative'
-  - 'disable_140chars'
-  - 'disable_class_inherits_from_params_class'
-  - 'disable_documentation'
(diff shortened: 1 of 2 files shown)


alexjfisher:
Re exclude_paths.  Why do we overwrite those configured by puppetlabs_spec_helper?

PR #574, opened 2019-07-01
Various fixes.

pillarsdotnet:
- Remove extra blank line at EOF of [`LICENSE`](https://github.com/voxpupuli/modulesync_config/commit/b51a9a12aa7a7b5ca062cecdf19718086ddd42b5)
- Add [`puppet-openvmtools`](https://github.com/voxpupuli/puppet-openvmtools) to [`managed_modules.yml`](https://github.com/voxpupuli/modulesync_config/commit/e94571e59c37deff17778097a0c6f73a57340217)
- Add missing langage specifier to code fence in [`moduleroot/.github/CONTRIBUTING.md.erb `](https://github.com/voxpupuli/modulesync_config/commit/71445da86be6f6b9c2db6b91698a0b21757e5688)
- Ensure newline at EOF of [`.yml` files](https://github.com/voxpupuli/modulesync_config/commit/ffdbf8aba28c684452072351ab6e4e6a7085424a).


moduleroot/.rubocop.yml.erb
```diff
@@ -1,6 +1,6 @@
 require: rubocop-rspec
 AllCops:
-  TargetRubyVersion: 1.9
+  TargetRubyVersion: 2.3
```
ekohl: In the past we had the question about whether we should stay compatible with older versions because they're also synced to older versions. Also unsure about JRuby compatibility levels. Anyone has something useful to say so we don't break anything?

pillarsdotnet: Rubocop 0.50 drops support for Ruby 1.9.  The oldest version it claims to support is 2.3.

ekohl: So the question is whether we need to stay syntax compatible with older rubies and I honestly don't know the answer. If we do, then we need to pin rubocop.

pillarsdotnet: ...But apparently Travis doesn't run Rubocop 0.50 ?  I'm getting different errors locally than I do via Travis-CI.

pillarsdotnet: Dropped the changes in `.rubocop.yml` since they conflict with the rubocop version preferred by vox-pupuli [`'~> 0.49.1'`](https://github.com/voxpupuli/modulesync_config/blob/master/config_defaults.yml#L52).

alexjfisher: jruby in puppetserver 5 is only ruby 1.9 compatible IIRC.

pillarsdotnet:
Breaking up into separate PR's per @alexjfisher .

PR #578, opened 2019-07-01
Ensure newline at EOF for .yml files.

pillarsdotnet:


pillarsdotnet:
@Dan33l Rebased and re-signed.

PR #583, opened 2019-07-24
Filters out the configured bundler path

logicminds:
Without this change simplecov shows reports for all the gems 
underneath the configured bundle path.


bastelfreak:
```
[1] pry(main)> require 'bundler'
=> true
[2] pry(main)> puts Bundler.configured_bundle_path.path
vendor/bundle/ruby/2.6.0
=> nil
[3] pry(main)>
```

PR #586, opened 2019-08-18
add puppet-lint-legacy_fact-check plugin

bastelfreak:
Depends on https://github.com/ekohl/puppet-lint-legacy_fact-check/issues/2


bastelfreak:
I switched this over to https://github.com/mmckinst/puppet-lint-legacy_facts-check as this seems to be more established. @ekohl objections?

ekohl:
No objections. Maybe I should remove my gem or at least point it to his to avoid confusion.

PR #587, opened 2019-09-11
add ability to see puppet debug log via env

foxxx0:
This makes debugging rspec tests a little easier by providing the
ability to capture the puppet debug log on stdout.
In order to use it, simply do:

```bash
DEBUG=true bundle exec rspec spec/classes/my_class_spec.rb
```


ekohl:
I'm wondering if this belongs in https://github.com/puppetlabs/puppetlabs_spec_helper so everyone can enjoy this.

PR #589, opened 2019-09-20
Use 3.9 as default facterlib facts version

alexjfisher:
DO NOT MERGE.  Just looking for feedback and/or alternative suggestions.

### DIFF
--- a/moduleroot/spec/spec_helper.rb.erb
+++ b/moduleroot/spec/spec_helper.rb.erb
@@ -60,8 +60,9 @@ end
 RSpec.configure do |c|
   <%- if @configs['hiera_config'] -%>
   c.hiera_config = <%= @configs['hiera_config'] %>
-
   <%- end -%>
+  c.default_facter_version = ENV['FACTERDB_FACTS_VERSION'] || '<%= @configs['facterdb_facts_version'] %>'
+
   # Coverage generation
   c.after(:suite) do
     RSpec::Puppet::Coverage.report!
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -37,7 +37,7 @@ Gemfile:
       - gem: puppetlabs_spec_helper
         version: '>= 2.14.0'
       - gem: rspec-puppet-facts
-        version: '>= 1.8.0'
+        version: '>= 1.9.5'
       - gem: rspec-puppet-utils
       - gem: puppet-lint-leading_zero-check
       - gem: puppet-lint-trailing_comma-check
@@ -130,5 +130,6 @@ spec/acceptance/nodesets/ubuntu-server-1604-x64.yml:
   delete: true
 spec/spec_helper.rb:
   mock_with: ':rspec'
+  facterdb_facts_version: '3.9'
 ...
 # vim: syntax=yaml


alexjfisher:
Or we could wait for https://tickets.puppetlabs.com/browse/PDK-1438
Not sure how practical that will be.  Perhaps someone at Puppet Inc is going to be keeping all the fact sets updated more regularly than currently happens.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -62,6 +62,12 @@ RSpec.configure do |c|
   c.hiera_config = <%= @configs['hiera_config'] %>
 
   <%- end -%>
+  case Puppet.version
+  when %r{^5}
+    c.default_facter_version = '3.9'
+  when %r{^6}
+    c.default_facter_version = '3.12'
```
alexjfisher: The first version to have shipped with puppet 6.  facterlib looks reasonable complete for this version too.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -62,6 +62,12 @@ RSpec.configure do |c|
   c.hiera_config = <%= @configs['hiera_config'] %>
 
   <%- end -%>
+  case Puppet.version
+  when %r{^5}
+    c.default_facter_version = '3.9'
```
alexjfisher: Random-ish version chosen here.  It shipped with Puppet 5.2 and 5.3.  It's the only 3.X version of facter with any archlinux facts in facterlib.

alexjfisher:
@ekohl I picked the 2 versions to use semi-randomly.  Do you think they're reasonable choices?  Given that not everyone uses AIO packages, there's no perfect answer.  Maybe just a single default would have been fine too.

Dan33l:
Even if distribution packagers can make different choices, following something near matrix version of puppet/facter that uses Puppet Inc. LGTM.
And the picked versions looks reasonable IMO.

moduleroot/spec/spec_helper.rb.erb
alexjfisher: Might change this to `3.13` just because there are ArchLinux facts for that version already.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -62,6 +62,8 @@ RSpec.configure do |c|
   c.hiera_config = <%= @configs['hiera_config'] %>
 
   <%- end -%>
+  c.default_facter_version = ENV['FACTERDB_FACTS_VERSION'] || '3.9'
```
ekohl: Perhaps we should also store this as a variable so it's easy to override:
```suggestion
  c.default_facter_version = ENV['FACTERDB_FACTS_VERSION'] || '<%= @configs['facterdb_facts_version'] %>'
```

alexjfisher: Updated.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -62,6 +62,8 @@ RSpec.configure do |c|
   c.hiera_config = <%= @configs['hiera_config'] %>
```
ekohl: Nit: IMHO this line can be dropped. Grouping the hiera config and default facter version together with trivial assignments is fine.

alexjfisher:
This has actually proved a little problematic.  If facterdb doesn't contain 3.9 facts for all OSes in metadata.json, rspec-puppet-facts hangs.  Discovered by @bastelfreak in the ipset module.
See https://github.com/mcanevet/rspec-puppet-facts/issues/100

PR #593, opened 2019-10-05
Add puppet-puppetwebhook

baurmatt:


bastelfreak:
Hi. It's not added to modulesync_config because it's managed by pdksync.

baurmatt:
Hey @bastelfreak, can you point me to the correct location? https://github.com/puppetlabs/pdksync/blob/master/managed_modules.yml doesn't have it listed.

Also VPT needs an update/fix because it shows puppet-puppetwebhook in `Really need an initial modulesync`.

PR #595, opened 2019-10-07
Fix-up CHANGELOG.md line endings

alexjfisher:
On Linux, workaround
https://github.com/github-changelog-generator/github-changelog-generator/issues/715
by appending to `changelog` task with code that performs the equivalent
of a `dos2unix`.

### DIFF
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -83,6 +83,19 @@ begin
     metadata = JSON.load(File.read(metadata_json))
     config.project = metadata['name']
   end
+
+  # Workaround for https://github.com/github-changelog-generator/github-changelog-generator/issues/715
+  require 'rbconfig'
+  if RbConfig::CONFIG['host_os'] =~ /linux/
+    task :changelog do
+      puts 'Fixing line endings...'
+      changelog_file = File.join(__dir__, 'CHANGELOG.md')
+      changelog_txt = File.read(changelog_file)
+      new_contents = changelog_txt.gsub(%r{\r\n}, "\n")
+      File.open(changelog_file, "w") {|file| file.puts new_contents }
+    end
+  end
+
 rescue LoadError
 end
 # vim: syntax=ruby


moduleroot/Rakefile.erb
```diff
@@ -83,6 +83,19 @@ begin
...
   end
+
+  # Workaround for https://github.com/github-changelog-generator/github-changelog-generator/issues/715
+  require 'rbconfig'
+  if RbConfig::CONFIG['host_os'] =~ /linux/
+    task :changelog do
+      puts 'Fixing line endings...'
+      changelog_file = File.join(File.dirname(__FILE__), 'CHANGELOG.md')
+      changelog_txt = File.read(changelog_file)
+      new_contents = changelog_txt.gsub(%r{\r\n}, "\n")
```
alexjfisher: This is the same gsub as the [stdlib function](https://github.com/puppetlabs/puppetlabs-stdlib/blob/5aa891c2ef8efce6eb6d1ff12e976b4753b3c4d9/lib/puppet/parser/functions/dos2unix.rb#L17).

moduleroot/Rakefile.erb
```diff
@@ -83,6 +83,19 @@ begin
     metadata = JSON.load(File.read(metadata_json))
     config.project = metadata['name']
   end
+
+  # Workaround for https://github.com/github-changelog-generator/github-changelog-generator/issues/715
+  require 'rbconfig'
+  if RbConfig::CONFIG['host_os'] =~ /linux/
+    task :changelog do
+      puts 'Fixing line endings...'
+      changelog_file = File.join(File.dirname(__FILE__), 'CHANGELOG.md')
```
ekohl: Slightly easier to read:
```suggestion
      changelog_file = File.join(__dir__, 'CHANGELOG.md')
```

ekohl:
Oh, in Foreman we hook into it by adding an extension:
https://github.com/theforeman/foreman-installer-modulesync/blob/38ff8bd86f347fbb686d8066680829898b5ec18f/moduleroot/Rakefile.erb#L18-L41

moduleroot/Rakefile.erb
```diff
@@ -83,6 +83,19 @@ begin
     metadata = JSON.load(File.read(metadata_json))
     config.project = metadata['name']
```
ekohl: Looking at this: `Blacksmith::Modulefile.new` should also provide this info and there's no need to load the metadata.json file again.

moduleroot/Rakefile.erb
alexjfisher: Updated.

moduleroot/Rakefile.erb
alexjfisher: I guess you were looking at this because it also uses `File.dirname(__FILE__)`? Since there might be a better solution, I've not updated here too.

ekohl: That's fine. I was looking at the task definition because Foreman does it that way.

moduleroot/Rakefile.erb
```diff
@@ -83,6 +83,19 @@ begin
...
   end
+
+  # Workaround for https://github.com/github-changelog-generator/github-changelog-generator/issues/715
+  require 'rbconfig'
+  if RbConfig::CONFIG['host_os'] =~ /linux/
+    task :changelog do
```
bastelfreak: Will this extend the current `changelog` task or overwrite it?

alexjfisher: extend it.

> [Appending to an existing rake task can be done by redefining it.](https://coderwall.com/p/qhdhgw/adding-a-post-execution-hook-to-the-rails-db-migrate-task)

PR #601, opened 2019-11-06
Add puppet-lint-anchor-check

alexjfisher:
Somehow forgot about adding this 2 years ago!

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -48,6 +48,7 @@ Gemfile:
       - gem: puppet-lint-absolute_classname-check
       - gem: puppet-lint-topscope-variable-check
       - gem: puppet-lint-legacy_facts-check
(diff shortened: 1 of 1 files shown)


ekohl:
Needs a rebase now. I do wonder if there's really no use for anchors anymore but I agree that in most cases it's not relevant anymore. Do you have any idea how many modules would be affected by this?

alexjfisher:
One module where it might still be needed, or at least difficult to refactor is puppet/selinux.  In the majority of cases `contain` should be used instead and for the others we identify, there's always the option to disable the check with a control-comment.

ekohl:
:+1: to using the control-comment.

bastelfreak:
@alexjfisher I cannot push to your branch. Please rebase.

PR #602, opened 2019-11-12
Use voxpupuli fixes branch of GHCG

ekohl:
These are no features in master compared to the release. Not using a git checkout saves having to always clone it, even when the release group is not being installed.

PR #604, opened 2019-11-28
Use the released version of GHCG

ekohl:
While this isn't perfect and still contains some bugs, there's currently nothing in master that's not also in the gem. Installation is quicker because it doesn't have to clone a git repository which Bundler always does regardless of the groups to figure out dependencies.


alexjfisher:
Duplicate of https://github.com/voxpupuli/modulesync_config/pull/602

PR #612, opened 2019-12-12
Use voxpupuli-test

ekohl:
This gem has most of the duplication that used to be modulesynced. By moving it to a gem, the need for modulesyncing is greatly reduced.

Currently a draft since we should do some bikeshedding and the gem should live in the Voxpupuli org.

Ideally we'd end up with by default a single line in `spec_helper.rb` that is `require 'voxpupuli/test/spec_helper'`. Whenever a module then needs overrides, it can be done by unmanaging the file rather than the complex .sync.yml dance that's done now.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -39,6 +39,7 @@ Gemfile:
   required:
     ':test':
       - gem: voxpupuli-test
+        version: '>= 1.0.0'
       - gem: coveralls
       - gem: simplecov-console
     ':development':
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -38,32 +38,9 @@
 Gemfile:
   required:
     ':test':
-      - gem: puppetlabs_spec_helper
-        version: '>= 2.14.0'
-      - gem: rspec-puppet-facts
-        version: '>= 1.9.5'
-      - gem: rspec-puppet-utils
-      - gem: puppet-lint-leading_zero-check
-      - gem: puppet-lint-trailing_comma-check
-      - gem: puppet-lint-version_comparison-check
-      - gem: puppet-lint-classes_and_types_beginning_with_digits-check
-      - gem: puppet-lint-unquoted_string-check
-      - gem: puppet-lint-variable_contains_upcase
-      - gem: puppet-lint-absolute_classname-check
-        version: '>= 2.0.0'
-      - gem: puppet-lint-topscope-variable-check
-      - gem: puppet-lint-legacy_facts-check
-      - gem: puppet-lint-anchor-check
-      - gem: metadata-json-lint
-      - gem: rubocop
-        version: '~> 0.49.1'
-      - gem: rubocop-rspec
-        version: '~> 1.15.0'
-      - gem: mocha
-        version: '~> 1.4.0'
+      - gem: voxpupuli-test
       - gem: coveralls
       - gem: simplecov-console
-      - gem: parallel_tests
     ':development':
       - gem: travis
       - gem: travis-lint
@@ -141,7 +118,7 @@ spec/acceptance/nodesets/windows-2012-serverstandard-x64.yml:
 spec/acceptance/nodesets/windows-2012R2-serverstandard-x64.yml:
   delete: true
 spec/spec_helper.rb:
-  mock_with: ':rspec'
+  mock_with: false
 CONTRIBUTING.md:
   delete: true
 ...
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -1,4 +1,4 @@
-require 'puppetlabs_spec_helper/rake_tasks'
+require 'voxpupuli/test/rake'
 
 # load optional tasks for releases
 # only available if gem group releases is installed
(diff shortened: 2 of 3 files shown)


moduleroot/Rakefile.erb
```diff
@@ -7,37 +7,6 @@ begin
 rescue LoadError
 end
 
-PuppetLint.configuration.log_format = '%{path}:%{line}:%{check}:%{KIND}:%{message}'
-
-desc 'Auto-correct puppet-lint offenses'
```
ekohl: I've entirely dropped this since the puppetlabs_spec_helper task `lint_fix` should be sufficient.

moduleroot/Rakefile.erb
```diff
@@ -7,37 +7,6 @@ begin
...
-task 'lint:auto_correct' do
-  Rake::Task[:lint_fix].invoke
-end
-
-desc 'Run acceptance tests'
```
ekohl: We normally use the `beaker` job and I'd like to standardize on that.

moduleroot/Rakefile.erb
```diff
@@ -7,37 +7,6 @@ begin
...
-RSpec::Core::RakeTask.new(:acceptance) do |t|
-  t.pattern = 'spec/acceptance'
-end
-
-desc 'Run tests'
```
ekohl: Also a customization that I've dropped for now.

moduleroot/Rakefile.erb
```diff
@@ -7,37 +7,6 @@ begin
...
-task test: [:release_checks]
-
-namespace :check do
```
ekohl: Moved to the gem

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -1,21 +1,19 @@
...
-require 'bundler'
-include RspecPuppetFacts
-
-if ENV['DEBUG']
-  Puppet::Util::Log.level = :debug
-  Puppet::Util::Log.newdestination(:console)
-end
+require 'voxpupuli/test/spec_helper'
 
 if File.exist?(File.join(__dir__, 'default_module_facts.yml'))
```
ekohl: I think that we can actually get rid of this since most have a small set. Once it's easy to unmanage the file, we should convert those to directly use `add_custom_fact` in those modules.

config_defaults.yml
```diff
@@ -34,33 +34,11 @@
...
-      - gem: puppet-lint-unquoted_string-check
-      - gem: puppet-lint-variable_contains_upcase
-      - gem: puppet-lint-absolute_classname-check
-        version: '>= 2.0.0'
-      - gem: puppet-lint-topscope-variable-check
-      - gem: puppet-lint-legacy_facts-check
-      - gem: puppet-lint-anchor-check
-      - gem: metadata-json-lint
+      - gem: voxpupuli-test
+        git: https://github.com/ekohl/voxpupuli-test-gem
```
ghoneycutt: Shouldn't this be in the voxpupuli namespace?

ekohl: See the second paragraph.

bastelfreak:
@ekohl I:
* release voxpupuli-test as 0.1.0
*  rebased your branch `vp-test-gem`
* proposed [a PR](https://github.com/voxpupuli/voxpupuli-test/pull/3) to voxpupuli-test to fix content we recently changed in modulesync_config

We we also add `coveralls` and `simplecov` as a runtime dependency to voxpupuli-test?

bastelfreak:
@ekohl I tested this with the master branch of voxpupuli-test and it seems to work fine. Do you still consider this WIP or should we release voxpupuli-test and afterwards merge this PR?

ekohl:
IMHO https://github.com/voxpupuli/voxpupuli-test/pull/4 should be merged and released. Also note I pushed another update after you did, but didn't test it so far. Other than that, I think this is ready for review.

config_defaults.yml
```diff
@@ -141,7 +118,7 @@ spec/acceptance/nodesets/windows-2012-serverstandard-x64.yml:
 spec/acceptance/nodesets/windows-2012R2-serverstandard-x64.yml:
   delete: true
 spec/spec_helper.rb:
-  mock_with: ':rspec'
+  mock_with: false
```
ekohl: Should we leave this out or keep it in as a reference on what to document?

alexjfisher:
Is it worth creating a few test PRs across some of our modules?

config_defaults.yml
bastelfreak: I think more documentation is always a good approach

bastelfreak:
@alexjfisher I'm currently releasing 0.2.0 of voxpupuli-test: https://github.com/voxpupuli/voxpupuli-test/pull/6
Afterwards we can test this on some modules.

ekohl:
This week I won't have that much time to work on this, but a few test PRs are certainly a good idea. Especially modules that use lib and do coverage

bastelfreak:
I tested this at https://github.com/voxpupuli/puppet-ferm/pull/94 and pushed one rubocop fix to this PR

bastelfreak:
I rebased this again and tested it on a few modules. I'm fine with merging it.

config_defaults.yml
```diff
@@ -38,32 +38,9 @@
...
-      - gem: puppet-lint-legacy_facts-check
-      - gem: puppet-lint-anchor-check
-      - gem: metadata-json-lint
-      - gem: rubocop
-        version: '~> 0.49.1'
-      - gem: rubocop-rspec
-        version: '~> 1.15.0'
-      - gem: mocha
-        version: '~> 1.4.0'
+      - gem: voxpupuli-test
```
ekohl: What are people's thoughts on pinning this? Should we release a 1.0.0 and stick to semver?

bastelfreak: since I didn't discover any further issues I'm :+1: for a 1.0.0 release + pinning to >= 1.0.0

config_defaults.yml
dhollinger: I'm a bit behind on this, why set `mock_with` to false?

ekohl: Because voxpupupli-test defaults to `:rspec` so there's no need to set it in the spec helper. This still allows an override but most modules shouldn't use it. This is in line with keeping the spec helper as minimal as possible. Ideally we'd standardize all our modules on rspec mocking and remove the need altogether, but we lack the resources to do so.

ekohl:
Since I'm the author of the PR, GH won't allow me to approve it. However, I think it's ready to merge.

PR #617, opened 2019-12-16
Skip command gem update

dhoppe:
This command is not necessary, because the gem ```rubygems-update``` will provide the executable ```bundle``` and the command ```gem update --system``` will make sure that it is up to date.

### DIFF
--- a/moduleroot/.travis.yml.erb
+++ b/moduleroot/.travis.yml.erb
@@ -22,7 +22,6 @@ addons:
 <%- end -%>
 before_install:
   - yes | gem update --system
-  - gem update bundler
   - bundle --version
 script:
   - 'bundle exec rake $CHECK'


ekohl:
Why do we even run `gem update --system`? Following https://github.com/puppetlabs/pdk-templates/pull/300 I think it's safe for us to completely drop it.

dhoppe:
@ekohl I am not sure, but some of the new features could be useful. For example that error messages are sent to STDERR:
* https://github.com/bundler/bundler/blob/v2.1.0/CHANGELOG.md#200pre1-2018-11-09

ekohl:
In theforeman modules we never had this line and I never missed it. On the other hand, we've had to update it multiple times in voxpupuli to fix issues. That's why I prefer to be on the safe side.

PR #631, opened 2020-01-10
Make 'before_install' in .travis.yml configurable

baurmatt:
This is needed to be able clone the puppetlabs-ruby_task_helper for
Bolt task spec tests.

See https://github.com/puppetlabs/puppetlabs-azure_inventory/blob/fee6f7e93559d5e72e3e6763a415883612f36a73/.travis.yml#L6 as an example of what I'd like to do.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -1,5 +1,8 @@
 ---
 .travis.yml:
+  before_install:
+    - yes | gem update --system
+    - bundle --version
   docker_sets:
   docker_defaults:
     # values will replace @@SET@@ with the docker_sets' value
--- a/moduleroot/.travis.yml.erb
+++ b/moduleroot/.travis.yml.erb
@@ -20,9 +20,12 @@ addons:
     <%- end -%>
   <%- end -%>
 <%- end -%>
+<%- if @configs['before_install'] -%>
 before_install:
-  - yes | gem update --system
-  - bundle --version
+<% @configs['before_install'].each do |bi| -%>
+  - <%= bi %>
+<% end -%>
+<%- end -%>
 script:
   - 'bundle exec rake $CHECK'
 matrix:


moduleroot/.travis.yml.erb
```diff
@@ -21,8 +21,9 @@ addons:
   <%- end -%>
 <%- end -%>
 before_install:
```
ekohl: Perhaps there should be a conditional before to skip this line if `before_install` is empty.

config_defaults.yml
```diff
@@ -1,5 +1,8 @@
 ---
 .travis.yml:
+  before_install:
+    - yes | gem update --system
```
ekohl: A bit unrelated to this, but I'm still of the opinion that we should drop this.

moduleroot/.travis.yml.erb
baurmatt: Thought about that while implementing, but other parameters in that file don't have one either. If you prefer to have a `if` block around it, I'm happy to add it.

config_defaults.yml
baurmatt: The `yes` part was introduced by https://github.com/voxpupuli/modulesync_config/pull/616

The `gem update --system` part seems to by originally introduced by https://github.com/voxpupuli/modulesync_config/pull/535

Should I open a followup issue for that?

bastelfreak: I think a separate PR for this is a good idea.

moduleroot/.travis.yml.erb
baurmatt: Added :)

config_defaults.yml
baurmatt: Opened https://github.com/voxpupuli/modulesync_config/issues/632 for that.

PR #633, opened 2020-01-17
Fix command to get latest tag

dhoppe:
The command ```git tag --list | tail -n 1``` does not point to the latest tag:
```
git tag --list | tail -n 1
2.9.0
```

This is related to the sorting of the tags:
```
git tag --list
...
2.0.0
2.1.0
2.10.0
2.10.1
2.2.0
2.3.0
2.4.0
2.5.0
2.6.0
2.7.0
2.8.0
2.9.0
```

The command ```git describe --abbrev=0``` will actually point to the latest tag:
```
git describe --abbrev=0
2.10.1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Module sync configurations for Vox Pupuli Modules
 ```bash
 git clone https://github.com/voxpupuli/modulesync_config.git
 cd modulesync_config
-git checkout $(git tag --list | tail -n 1) # checkout latest tag
+git checkout $(git tag --list | sort -V | tail -n1) # checkout latest tag
 bundle install
 bundle exec msync help update
 ```


dhoppe:
Hm. It looks like this command only works on annotated tags. :(

bastelfreak:
Since we only have annotated tags, this would be fine?

dhoppe:
No. I did a loop over all repositories and found several where the new command did not work. For example puppet-zabbix.

ghoneycutt:
Looked and non-annotated tags are pervasive. Sort has a way of handling version numbers.

`git tag --list | sort -V | tail -n1`

ghoneycutt:
Simple test

```sh
$ git tag -l
1
2
2.10.1
2.9.0

$ git tag --list |  tail -n1
2.9.0

$ git tag --list | sort -V | tail -n1
2.10.1
```

dhoppe:
@ghoneycutt Thx. I just updated my commit.

PR #638, opened 2020-02-09
Drop `redcarpet` gem

bastelfreak:
Initially introduced in da540ad5f56192c07f2e764c484431264b62fe7b as a
missing dependency to puppet-strings. I tested this on the puppet-borg
module. The gem isn't required to build our REFERENCE.md.


alexjfisher:
I guess if we're mistaken, we can always add it back.

PR #641, opened 2020-02-16
Ignore trailing whitespace in CHANGELOG.md since it is auto generated.

ghoneycutt:


alexjfisher:
FWIW, the rake task already handles this and this should only be a problem if running github-changelog-generator by hand.

PR #646, opened 2020-04-05
Set hiera_config after including the spec helper

ekohl:
Setting hiera_config is only valid after loading rspec-puppet.


ekohl:
This affects two modulesyncs and I've updated both: https://github.com/voxpupuli/puppet-telegraf/pull/140 & https://github.com/voxpupuli/puppet-openvpn/pull/376.

PR #649, opened 2020-04-19
spec_helper_acceptance: Add helper to load custom files

bastelfreak:
This allows us to keept the spec_helper_acceptance clean and managed by
modulesync, but it also provides enough flexibility for our module
developers to inject custom code into the test suite.

Also this alligns our setup with the one from the foreman people.


bastelfreak:
This was tested in https://github.com/voxpupuli/puppet-zabbix/pull/682

PR #654, opened 2020-05-12
Manage GitHub settings via Probot

dhoppe:
This pull request allows us to manage the GitHub settings via Probot. I left out the labels, as they are managed by ```vox-pupuli-tasks```. 

Maybe we can also read the file ```metadata.json``` to automatically add values for ```name```, ```description```, ```homepage``` and ```topics```. 

Unfortunately, the ```delete_merge_on_branch``` parameter does not work yet: 
* https://github.com/probot/settings/issues/134

Also, the rule for the ```modulesync``` branch will not be applied if it does not already exist:
* https://github.com/probot/settings/issues/136

### DIFF
--- a/moduleroot/.github/settings.yml
+++ b/moduleroot/.github/settings.yml
@@ -1,0 +1,36 @@
+repository:
+  private: false
+  has_issues: true
+  has_projects: false
+  has_wiki: false
+  has_downloads: true
+  default_branch: master
+  allow_squash_merge: true
+  allow_merge_commit: true
+  allow_rebase_merge: true
+  delete_branch_on_merge: true
+  archived: false
+
+branches:
+  - name: master
+    protection:
+      required_pull_request_reviews:
+        required_approving_review_count: 1
+        dismiss_stale_reviews: true
+        require_code_owner_reviews: true
+      required_status_checks:
+        strict: true
+        contexts: ['continuous-integration/travis-ci']
+      enforce_admins: true
+      restrictions: null
+      required_signatures: true
+  - name: modulesync
+    protection:
+      required_pull_request_reviews: null
+      required_status_checks:
+        strict: true
+        contexts: ['continuous-integration/travis-ci']
+      enforce_admins: true
+      restrictions: null
+      required_signatures: true
+
--- a/moduleroot/CODEOWNERS
+++ b/moduleroot/CODEOWNERS
@@ -1,0 +1,3 @@
+.github/settings.yml @voxpupuli/project-maintainers
+
+* @voxpupuli/collaborators


moduleroot/.github/settings.yml
```diff
@@ -0,0 +1,36 @@
...
+branches:
+  - name: master
+    protection:
+      required_pull_request_reviews:
+        required_approving_review_count: 1
+        dismiss_stale_reviews: true
+        require_code_owner_reviews: true
+      required_status_checks:
+        strict: true
+        contexts: ['continuous-integration/travis-ci']
```
ekohl, line 23: I think this doesn't work for our release workflow. Also not sure if there might be a difference between legacy open source Travis and the modern one. Might be that we have some modules using the modern one.

moduleroot/.github/settings.yml
```diff
@@ -0,0 +1,36 @@
+repository:
+  private: false
+  has_issues: true
+  has_projects: false
+  has_wiki: false
```
ekohl, line 5: I like this, but have we checked whether this is actually true? Or do you want to do that as part of the modulesync review?

moduleroot/.github/settings.yml
dhoppe: @bastelfreak mentioned that the ```travis_relase``` task might fail, but we could set ```enforce_admins: false``` to override this behavior. 

I do not know, if we are able to check the current Travis CI version, but it should be able to configure multiple contexts like Travis CI and Jenkins. In the end it is an array, but I have to take a look at the docs, if this will be treated as ```and``` condition.

moduleroot/.github/settings.yml
dhoppe: It should be possible to loop over ```managed_modules.yml``` to fetch the current values regarding ```projects``` and ```wiki```.

ekohl:
One thing to consider is that there are people who use our modulesync on their own modules. On the other hand, they can delete the file if they want to.

dhoppe:
We should add a section to the ```README.md``` that these two files need to be deleted via ```.sync.yml``` or modified. To be honest, we might need to convert them to ```.erb```. This allows other projects to modify them or define excludes for a set of repositories.

moduleroot/.github/settings.yml
bastelfreak: as far as I know this will break our release process (that's why I tested this long time ago and had to remove the branch protection again). This will enforce that travis runs for every PR and only PRs are allowed for the protected branch. You cannot push anymore directly to master. This in general should be our goal, but our travis_release task directly pushes to master:
https://github.com/voxpupuli/voxpupuli-release-gem/blob/fb73087fda1f64b79ee3bb668e3a4ce6cd8626b1/lib/voxpupuli/release/rake_tasks.rb#L32-L33

If I remember corretly nobody can push to master, even admins. admins "only" have the power to merge PRs if travis is still running/failing.

dhoppe: @bastelfreak I think you have missed my comment regarding ```enforce_admins: null```. 

I was able to merge a pull request, although the conditions for the Travis CI check were not fulfilled:
* https://github.com/dhoppe/foo/pull/1

I also was able to push to the master branch:
* https://github.com/dhoppe/foo/commit/8ef29acccdd6d1b94dfb3a59ee60e4bab96f894d

This might be related to the ```CODEOWNERS```.

I think it would be best if I make a template from the files and we start with some small changes in the ```config_defaults.yml``` file.

moduleroot/.github/settings.yml.erb
```diff
@@ -0,0 +1,52 @@
...
+  - name: <%= label['name'] %>
+<%     label.keys.each do |k| -%>
+<%       next if k == 'name' -%>
+    <%= k %>: <%= label[k] %>
+<%     end -%>
+<%   end -%>
+<% end -%>
+<% if @configs['branches'] -%>
+
+branches:
```
dhoppe, line 20: I am pretty sure that this can be done easier.

moduleroot/.github/settings.yml
ekohl: It has to be tested by Travis, but Travis/Github identifies by commit. That means you can push it to a branch, get it tested and merged. Perhaps you can change the script to push to a `next-release` branch and set up auto merge rules for that with something like [mergify](https://mergify.io/)

moduleroot/.github/settings.yml.erb
```diff
@@ -0,0 +1,52 @@
...
+<%   @configs['branches'].each do |branch| -%>
+  - name: <%= branch['name'] %>
+<%     branch.keys.each do |k| -%>
```
ekohl, line 23: You can also iterate over the tuples
```suggestion
<%     branch.keys.each do |key, value| -%>
```

This avoids the whole `branch[k]` part.

moduleroot/.github/settings.yml.erb
```diff
@@ -0,0 +1,52 @@
...
+<%       next if k == 'name' -%>
+    <%= k %>:
+<%       if branch[k].has_key?('required_pull_request_reviews') -%>
+<%         if branch[k]['required_pull_request_reviews'].nil? -%>
```
ekohl, line 27: `branch[k]['required_pull_request_reviews']` is nil, whether the key exists or not. It's not like Python which raises an exception if the key is not set.

moduleroot/.github/settings.yml.erb
```diff
@@ -0,0 +1,52 @@
...
+<% if @configs['labels'] -%>
+
+labels:
+<%   @configs['labels'].each do |label| -%>
+  - name: <%= label['name'] %>
+<%     label.keys.each do |k| -%>
+<%       next if k == 'name' -%>
+    <%= k %>: <%= label[k] %>
+<%     end -%>
+<%   end -%>
```
ekohl, line 16: Not sure if delete is pretty, but I think it works:
```suggestion
  - name: <%= label.delete('name') %>
<%     label.each do |key, value| -%>
    <%= key %>: <%= value %>
<%     end -%>
<%   end -%>
```

moduleroot/.github/settings.yml.erb
```diff
@@ -0,0 +1,52 @@
+<% if @configs['repository'] -%>
```
ekohl, line 1: Another thought for this file:
```ruby
data = {}
['repository', 'labels', 'branches'].each do |key|
  data[key] = @configs[key] if @configs[key]
end
-%>
<%= data.to_yaml %>
```

moduleroot/.github/settings.yml
dhoppe: I will remove the ```required_pull_request_reviews```, ```required_status_checks```, ```enforce_admins``` and ```restrictions``` to get started.

Thanks for the tip on ```mergify```. I will ll take a look at it later.

moduleroot/.github/settings.yml.erb
dhoppe: That is what I was looking for, but it was so late. However ```~``` is not interpreted as ```null```.

moduleroot/.github/settings.yml
dhoppe: Many repositories use ```projects```and / or ```wiki```because this is a default value during the creation process. I think we need to check this manually during the ```modulesync``` and modify it if necessary via ```.sync.yml```.

moduleroot/CODEOWNERS.erb
```diff
@@ -0,0 +1,5 @@
+<% if @configs['permissions'] -%>
+<%   @configs['permissions'].each do |key, value| -%>
+<%= key %> <%= value %>
+<%   end -%>
+<% end -%>
```
ekohl, line 5: Untested, but I think this will result in the same:
```suggestion
<% @configs['permissions']&.each do |key, value| -%>
<%= key %> <%= value %>
<% end -%>
```

dhoppe:
I think now we have a working version that we could test on a repository. 

This should cover most of the features of https://github.com/voxpupuli/vox-pupuli-tasks/issues/14

moduleroot/CODEOWNERS.erb
dhoppe: It does. Honestly, how many Ruby books do you have hidden under your pillow? ;)

ekohl: None, I still consider myself a novice Ruby developer tbh. Just look at a bunch of other devs  and PR reviews. Copy that. That's how I learned the language constructs. This one is called the [safe navigation operator](https://en.wikipedia.org/wiki/Safe_navigation_operator#Ruby) and was introduced in Ruby 2.3.

Also read the core data structures ([Array](https://ruby-doc.org/core-2.7.1/Array.html), [Hash](https://ruby-doc.org/core-2.7.1/Hash.html), [String](https://ruby-doc.org/core-2.7.1/String.html) and [Integer](https://ruby-doc.org/core-2.7.1/Integer.html)) when you deal with them. The list of methods is useful.

Once you have a decent understanding of what's possible, you can express yourself much more powerful.

config_defaults.yml
```diff
@@ -35,6 +70,10 @@
     channels:
     - "chat.freenode.org#voxpupuli-notifications"
   user: puppet
+CODEOWNERS:
+  permissions:
+    '.github/settings.yml': '@voxpupuli/project-maintainers'
+    '*': '@voxpupuli/collaborators'
```
bastelfreak, line 76: this will send an email to every collaborator, right? Is that maybe too noisy?

ekohl: Right now I purposely unwatched a lot of repositories precisely because of it.

dhoppe: Damn. This was supposed to be a failsafe, according to the docs of Probot. I will drop the last line, but the first one is necessary:
* https://github.com/probot/settings#security-implications

dhoppe:
I think I understand the problem now. GitHub allows things on its website that do not work with Curl or Probot. 

I actually think it is quite nice when a review of the pull request is needed. However, this does not work with the task ```travis_release```. Therefore I made two changes. 

However, I will still contact GitHub support to have branches automatically deleted after a merge and to create protection rules for non-existent branches.

config_defaults.yml
```diff
@@ -1,4 +1,37 @@
 ---
+.github/settings.yml:
+  repository:
+    private: false
+    has_issues: true
+    has_projects: false
+    has_wiki: false
+    has_downloads: true
+    default_branch: master
```
wyardley, line 9: Maybe out of scope for this specific PR, but has there been discussion of changing the default branch name to main?

ekohl: AFAIK there has not been any discussion on this.

PR #658, opened 2020-07-11
Disable RSpec/MultipleExpectations

alexjfisher:
This would allow us to upgrade rubocop-rspec to 1.16 without creating a
whole bunch of extra work.

See https://github.com/voxpupuli/voxpupuli-test/pull/26

### DIFF
--- a/moduleroot/.rubocop.yml.erb
+++ b/moduleroot/.rubocop.yml.erb
@@ -533,6 +533,9 @@ RSpec/RepeatedDescription:
 RSpec/NestedGroups:
   Enabled: False
 
+RSpec/MultipleExpectations:
+  Enabled: false
+
 # this is broken on ruby1.9
 Layout/IndentHeredoc:
   Enabled: False


igalic:
it would be nice if `rubocop --auto-fix` would fix that…

alexjfisher:
> In general I think this cop is a nice idea, but multiple expectations have their uses. For example, it's much faster if you don't need to compile the entire catalog again. AFAIK `it { }` does compile it every time.

... not sure.  Maybe not?? I'm not clever enough to understand how rspec puppet catalog caching works. I remember someone looking into this in the past though... https://github.com/rodjek/rspec-puppet/pull/329 ??? ;)

alexjfisher:
I'm not against the cop per se either. Just at this particular moment didn't want to go create a whole lot of yak shaving just to solve https://github.com/voxpupuli/pdksync_config/issues/1#issuecomment-613550313

PR #660, opened 2020-07-24
require voxpupuli-test 2 or newer

bastelfreak:


config_defaults.yml
```diff
@@ -39,7 +39,7 @@ Gemfile:
   required:
     ':test':
       - gem: voxpupuli-test
-        version: '>= 1.4.0'
+        version: '>= 2.0.0'
```
ekohl, line 42: Something I think that was wrong in the past is that we didn't respect semver. That means a lot of our modules are now failing
```suggestion
        version: '~> 2.0'
```

bastelfreak:
@ekohl can you approve now?

PR #666, opened 2020-08-12
Enable Security/YAMLLoad cop

ekohl:
Luckily we no longer have to care about Ruby 2.0 development environments and using safe_load is a good practice.


alexjfisher:
Our target version is still 1.9 (due to Puppetserver 5).  Is that an issue? (or was it specifically just a 2.0.0 workaround?)

ekohl:
Is it 1.9 syntax or 1.9 runtime? I thought it was just the language level but now I'm wondering.

ekohl:
Closing since https://github.com/voxpupuli/modulesync_config/pull/667 was merged.

PR #667, opened 2020-08-12
Use Rubocop config from voxpupuli-test

ekohl:
This is still a draft since it depends on https://github.com/voxpupuli/voxpupuli-test/pull/28 and assumes that will be released as version 2.1.0.


moduleroot/.rubocop.yml.erb
```diff
@@ -1,554 +1,2 @@
...
-# This affects hiera interpolation, as well as some configs that we push.
-Style/FormatStringToken:
-  Enabled: false
-
-# This is useful, but sometimes a little too picky about where unit tests files
-# are located.
-RSpec/FilePath:
-  Enabled: false
+inherit_gem:
+  voxpupuli-test: rubocop.yml
```
alexjfisher, line 2: Should we include a `inherit_from: .rubocop_todo.yml` as well?

ekohl: None of our modules have this today and if the file doesn't exist, Rubocop hard crashes with `No such file or directory`.

alexjfisher: Oh, that’s annoying. Since it’s still erb, can It dynamically add the line if the file exists?
I’m thinking it might be useful when we finally drop ruby 1.9 support at the end of the year.

ekohl: Technically yes and the recent support in modulesync does make that easier (since you get a variable for the target directory).

ekohl:
Now that 2.1 has been released, this is ready for review. I'd prefer to leave out the TODO file since that's not really used in our modules and I'd like to avoid encouraging it. That's also a nice way of saying that I'm too lazy to write the code for it ;)

bastelfreak (approved):

moduleroot/.rubocop.yml.erb
alexjfisher: We can always add this later.

alexjfisher:
> Now that 2.1 has been released, this is ready for review. I'd prefer to leave out the TODO file since that's not really used in our modules and I'd like to avoid encouraging it. That's also a nice way of saying that I'm too lazy to write the code for it ;)

I think it'll be useful when/if we finally bump to much newer versions of rubocop after we drop ruby 1.9 support.  But we don't need it right now, so +1 from me.

alexjfisher (approved):

PR #669, opened 2020-08-15
dont manage mcollective module

bastelfreak:
the module is deprecated, choria-io maintains up2date versions.

### DIFF
--- a/managed_modules.yml
+++ b/managed_modules.yml
@@ -60,7 +60,6 @@
 - puppet-lldpd
 - puppet-logrotate
 - puppet-make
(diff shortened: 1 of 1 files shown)


ekohl:
https://github.com/voxpupuli/puppet-mcollective doesn't mention that it's deprecated. Is that something we want to do or should users know that Choria is the new MCO?

bastelfreak:
@ekohl I think we should finish https://github.com/voxpupuli/voxpupuli.github.io/pull/213 and define how we can officially deprecate/archive a module.

PR #674, opened 2020-09-23
Use corresponding ruby versions for the Puppet versions while testing

ghoneycutt:
Current Puppet 6 uses ruby 2.5.7 and Puppet 5 uses 2.4.9.

https://puppet.com/docs/puppet/latest/about_agent.html
https://puppet.com/docs/puppet/5.5/about_agent.html

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -7,25 +7,25 @@
   docker_sets:
   docker_defaults:
     # values will replace @@SET@@ with the docker_sets' value
-    - rvm: 2.5.3
+    - rvm: 2.5.7
       services: docker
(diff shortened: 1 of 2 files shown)


igalic (approved):

ekohl (commented):
I still think we should just use `2.5` and `2.4` without the z-part. Keeping it in sync is really of minimal value IMHO since I strongly doubt we'll find real issues due to mismatches. If you want real tests, there are acceptance tests with the actual versions. Not keeping exact versions also keeps your git logs a little bit cleaner.

ghoneycutt:
I use the same ruby versions to minimize any variance between the testing framework and the actual versions used. By not specifying the z-part, you open yourself up to drift. Yes you occasionally have to update the numbers, but that's software :)

ekohl:
We no longer maintain Ruby versions at all in modulesync.

PR #680, opened 2020-10-30
Don't ship `spec` directory, `Rakefile` or `Gemfile` in packaged modules

alexjfisher:
This should make the average module a fair bit smaller.
PDK already does this by default.
https://github.com/puppetlabs/pdk-templates/blob/172738acbf55e65afb1eb8541deb8bb242ca877b/config_defaults.yml#L57

### DIFF
--- a/moduleroot/.pmtignore.erb
+++ b/moduleroot/.pmtignore.erb
@@ -1,11 +1,12 @@
 docs/
 pkg/
+Gemfile
 Gemfile.lock
 Gemfile.local
 vendor/
 .vendor/
-spec/fixtures/manifests/
-spec/fixtures/modules/
+spec/
+Rakefile
 .vagrant/
 .bundle/
 .ruby-version


alexjfisher:
It might look innocent enough, but this is quite a big change.  Even if you think this is a great idea, please don't merge just yet.  But please leave feedback.

ekohl (approved):
In Foreman's modulesync we do the same:
https://github.com/theforeman/foreman-installer-modulesync/blob/8e29149d17efd144371d5c46baff9f04fbb79141/moduleroot/.pmtignore.erb#L51
Haven't heard any complaints.

You may also want to drop the Gemfile and Rakefile.

alexjfisher:
> In Foreman's modulesync we do the same:
> https://github.com/theforeman/foreman-installer-modulesync/blob/8e29149d17efd144371d5c46baff9f04fbb79141/moduleroot/.pmtignore.erb#L51
> Haven't heard any complaints.
> 
> You may also want to drop the Gemfile and Rakefile.

Good plan.  This also matches the PDK behaviour.

ekohl (approved):

ekohl:
Would you mind rebasing the PR?

ekohl (approved):

Issue #428, opened 2017-11-18
Run acceptance tests against all supported operatingsystems


ekohl:
* [ ] https://github.com/voxpupuli/puppet-autofs/pull/85
* [ ] https://github.com/voxpupuli/puppet-collectd/pull/732
* [ ] https://github.com/voxpupuli/puppet-corosync/pull/426
* [ ] https://github.com/voxpupuli/puppet-gluster/pull/144
* [x] https://github.com/voxpupuli/puppet-nscd/pull/24
* [x] https://github.com/voxpupuli/puppet-report_hipchat/pull/32
* [x] https://github.com/voxpupuli/puppet-yum/pull/79

bastelfreak:
I just had a discussion with @baurmatt about this. Maybe it would be cool to check all OSes in the metadata.json in an msync run and update the .sync.yml / .travis.yml with suitable beaker jobs.

dhoppe:
This could be difficult, because there are a lot of metadata.json which do not provide a version for ArchLinux or FreeBSD etc.

baurmatt:
So how about a whitelist for OS/Version which are, in general, known to work? Special OS like Arch, Gentoo, FreeBSD, Windows, Raspbian need to be managed manually until we figure something out for them.

ekohl:
https://github.com/voxpupuli/modulesync/pull/168 should allow looking at metadata.json. We can then write a method that automatically maps things and skips unknown ones.

ekohl:
This is now done.

PR #683, opened 2020-11-06
Switch from Travis CI to Github Actions

ekohl:
* [ ] Modules with special care:
  * [ ] https://github.com/voxpupuli/puppet-vault_lookup/pull/33 tests are broken because changes are redacted
  * [x] https://github.com/voxpupuli/puppet-spiped/pull/23
  * [ ] https://github.com/voxpupuli/puppet-gitlab_ci_runner/pull/93 Installs ruby_task_helper via git and calls scripts/start-gitlab.sh in before_install. The helper looks more like a fixture. Starting gitlab for beaker tests requires manual attention.
  * [x] https://github.com/voxpupuli/puppet-rundeck/pull/455
* [x] Verify release workflow
  * [x] Decide to run tests or not


moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -0,0 +1,139 @@
...
+      - name: Install dependencies
+        run: |
+          gem install bundler
+          bundle config path vendor/bundle
+          bundle config without 'development system_tests release'
+          bundle install --jobs $(nproc) --retry 3
+      - name: Run tests
+        run: bundle exec rake
+<%- if Dir[File.join(@metadata[:workdir], 'spec', 'acceptance', '**', '*_spec.rb')].any? -%>
+
```
bastelfreak, line 46: is the empty line on purpose?

bastelfreak:
did you test this on a module already? maybe on https://github.com/voxpupuli/puppet-bird/tree/master/.github/workflows since it already has actions for tests. I'm also fine with running the tests on github but keeping the release on travis for now. I'm happy to  convert the release task later.

moduleroot/.github/workflows/ci.yml.erb
ekohl: Yes. That way you get an empty line between unit tests and acceptance tests.

ekohl:
Not yet, but I've taken https://github.com/theforeman/foreman-installer-modulesync/blob/master/moduleroot/.github/workflows/cron.yml.erb with some changes that I already have PRs for and modified to be a single file. In hindsight we should have probably done the same thing with Foreman. This was just a code dump of my WIP before the weekend.

The next steps would be to see what the various modules need. One big difference is that rather than statically determining beaker sets this all detects it dynamic (fixing https://github.com/voxpupuli/modulesync_config/issues/428). This works in the Foreman ecosystem but won't always work here. For example, I already know that puppet-selinux has these tests but can't run in Docker.

moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -0,0 +1,139 @@
...
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-2.7-gems-acceptance-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-2.7-gems-acceptance-
+      - name: Install dependencies
+        run: |
+          gem install bundler
+          bundle config path vendor/bundle
+          bundle config without 'development test release'
```
bastelfreak, line 129: our acceptance tests require at least the rake gem but don't pull it on. On our travis tests, we install system_tests and test groups for acceptance tests. I'm not sure if we should add rake as a dependency to voxpupuli-acceptance or install the test group.

ekohl: Good point. I did that with Foreman's modules by adding Rake as an ungrouped dependency. This made sense to me because essentially all workflows go through rake tasks.

bastelfreak: sounds good to me!

moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -0,0 +1,139 @@
...
+        <%= option %>:
+        <%- values.each do |value| -%>
+          - "<%= value %>"
+        <%- end -%>
+        <%- end -%>
+        <%- if @configs['excludes'].any? -%>
+        exclude:
+        <%- @configs['excludes'].each do |exclude| -%>
+        <%- exclude.each do |key, value| -%>
+          <%= key == exclude.first.first ? '-' : ' ' %> <%= key %>: "<%= value %>"
```
bastelfreak, line 98: can you explain the syntax a bit? I think we've cases where we want to exclude an OS completely (because beaker doesn't support it) or puppet doesn't provide proper packages (e.g. Ubuntu 20.04 only supports Puppet 6).

ekohl: https://github.com/theforeman/puppet-puppet/blob/635380f7497700d3cb0a992805ff044684f2c3a4/.sync.yml#L2-L7

moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -0,0 +1,139 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
+          - "2.5"
+          - "2.4"
+        puppet:
+          - "6"
+          - "5"
```
bastelfreak, line 16: since 5 is almost EOL, I think we don't need to continue testing on it

ekohl:
Right now it respects `metadata.json`'s Puppet version requirement. That means you can drop Puppet 5.x in `metadata.json` and it'll stop testing it. It also automatically tests all supported OS versions with acceptance tests if they're present. That means in theory we can add Puppet 7 to `metadata.json` and it should start testing it.

See https://github.com/voxpupuli/puppet-borg/pull/78 for an example.

It still needs to fix the `voxpupuli-test` dependency. Right now the `Rakefile` hard requires it.

Then we also need a replacement for the release workflow.

moduleroot/Rakefile.erb
```diff
@@ -18,14 +27,17 @@ task test_with_coveralls: [:test] do
   end
 end
 
-desc 'Generate REFERENCE.md'
-task :reference, [:debug, :backtrace] do |t, args|
-  patterns = '<%= @configs['puppet_strings_patterns'].join(" ") %>'
-  Rake::Task['strings:generate:reference'].invoke(patterns, args[:debug], args[:backtrace])
+if Rake::Task.task_defined?('strings:generate:reference')
```
ekohl, line 30: We only used this in one place but that's gone now - should we drop it? @alexjfisher you altered the invocation in extlib for deprecated ones. Thoughts?

ekohl:
So short overview of overrides for Travis:
* puppet-vault_lookup/.sync.yml Has a complex matrix but that should be redundant, especially after GHA migration
* puppet-spiped/.sync.yml Doesn't manage .travis.yml because it does beaker testing with multiple hosts. The Github Actions PR would also break here. Requires manual attention.
* puppet-gitlab_ci_runner/.sync.yml Installs ruby_task_helper via git and calls `scripts/start-gitlab.sh` in before_install. The helper looks more like a fixture. Starting gitlab for beaker tests requires manual attention.
* puppet-rundeck/.sync.yml Installs libaugeas-dev. Probably still needed

moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -0,0 +1,100 @@
+name: CI
```
bastelfreak, line 1: does it make sense to add a comment like  `# THIS FILE COMES FROM https://github.com/voxpupuli/modulesync_config`?

Flipez:
Regarding the mail about the REFERENCE.md: We currently check this file in VPT and display an error/warning if it is missing.

We do not really depend on this file - so if it will be no longer generated we can simply stop checking it, but I assume that it has some sort of content that others might depend on?

ekohl:
> I assume that it has some sort of content that others might depend on?

Compare https://forge.puppet.com/modules/ekohl/example/reference vs https://forge.puppet.com/modules/puppet/example. The former included a REFERENCE.md file, the latter didn't. For the forge it only matters what's in the release tarball, so you can (and in Foreman we do) generate it only prior to releasing and add it to `.gitignore`. This means you don't have any outdated file in git but with the downside that you can't go to Github to read the documentation. Personally I think that's fine but I know opinions differ on this.

ekohl:
The current Travis release runs the entire test matrix. However, it only uses the unit tests on one axis to determine if it should release. The GHA flow doesn't run any tests. My reasoning is that we already run the tests on the PR. A release should include all PRs (due to the changelog) so running tests is redundant. The benefit is that releases are faster (less time to wait) and less chance something fails (intermittent network failure).

With that said, I think we should merge the current verison and do syncs.

https://github.com/voxpupuli/puppet-example/pull/2 is an example of how the new dynamic matrix makes adding a new Puppet version trivial. No more test matrix updates, just stick to metadata.json.

ekohl:
I've now spent most of my week on this and I think we should pull the trigger. There may be some modules that have issues (2 noted in the TODO).

The big question remains: do we run tests on release or not?

bastelfreak (approved):
IMO it's fine to not run the tests. I think we should merge this as is, convert as many modules as possible and check the remaining modules later on. I can do a modulesync_config release with this PR and kick off msync later on.

PR #685, opened 2020-11-18
Extend .pmtignore to trim modules

ekohl:
This follows c294770cb0b2608cc498fd7aeace60a2888b2106 which dropped specs. It now also drops all hidden files that serve no purpose on production environments.

### DIFF
--- a/moduleroot/.pmtignore.erb
+++ b/moduleroot/.pmtignore.erb
@@ -14,11 +14,23 @@ coverage/
 log/
 .idea/
 .dependencies/
+.github/
 .librarian/
 Puppetfile.lock
 *.iml
+.editorconfig
+.fixtures.yml
+.gitignore
+.msync.yml
+.overcommit.yml
+.pmtignore
+.rspec
+.rspec_parallel
+.rubocop.yml
+.sync.yml
 .*.sw?
 .yardoc/
+.yardopts
 Dockerfile
 <% if ! @configs['paths'].nil? -%>
 <% @configs['paths'].each do |path| -%>


bastelfreak (approved):

moduleroot/.pmtignore.erb
```diff
@@ -14,11 +14,23 @@ coverage/
...
+.github/
 .librarian/
 Puppetfile.lock
 *.iml
+.editorconfig
+.fixtures.yml
+.gitignore
+.msync.yml
+.overcommit.yml
+.pmtignore
```
igalic, line 26: i don't know why, but this is hilarious to me

igalic (approved):

alexjfisher:
I'm pretty sure hidden files are not packaged by default?

ekohl:
> I'm pretty sure hidden files are not packaged by default?

They are. Perhaps this is a change that should be submitted to [puppet-modulebuilder](https://github.com/puppetlabs/puppet-modulebuilder)?

alexjfisher:
> > I'm pretty sure hidden files are not packaged by default?
> 
> They are. Perhaps this is a change that should be submitted to [puppet-modulebuilder](https://github.com/puppetlabs/puppet-modulebuilder)?

I think so, as it differs from how `puppet module build` behaved.

alexjfisher:
PMT excluded these files by default
https://github.com/puppetlabs/puppet/blob/5baab1e14226d284c3251ae34b713e6580de358f/lib/puppet/module_tool.rb#L13
https://github.com/puppetlabs/puppet/blob/5baab1e14226d284c3251ae34b713e6580de358f/lib/puppet/module_tool/applications/builder.rb#L98

alexjfisher (approved):

ekohl:
Note that PDK ignores the same files: https://github.com/voxpupuli/pdk-templates/blob/bb554cbb831d8c651cd54056ac09950228603e5e/config_defaults.yml#L40-L57

PR #686, opened 2020-11-25
release 4.0.0

bastelfreak:


ekohl (approved):

ekohl:
Note that when doing these, .sync.yml should also be cleaned up from any `.travis.yml` overrides. Perhaps it can be scripted?

PR #690, opened 2020-12-08
Run rubocop checks in matrix build

traylenator:
Run `bundle exec rake rubocop`.

This requires we install voxpupuli-test to get the rubocop
dependencies.


ekohl (commented):
I was thinking about the same thing. Perhaps we should also run `lint` before the entire matrix? That means we only run a minimal CI if you have obvious blockers to merging.

traylenator:
Will save a lot of docker images for sure.

ekohl:
Btw, I opened https://github.com/puppetlabs/puppetlabs_spec_helper/pull/327 which would also make it easy to run all pre-release checks during the validation. I also started a PR to voxpupuli-test, but then first wanted to complete https://github.com/voxpupuli/voxpupuli-test/pull/37. Then the actual unit tests can be updated to run `rake parallel_spec` instead of the default task.

ekohl (approved):

PR #692, opened 2020-12-14
Introduce default job length of 40 minutes

traylenator:
Override default job timeout for Github actions.

Set the job timeout to 40 minutes rather than the default
360 minutes (6 hours)

Running some test jobs with 1 minute the queued time per job does not
appear to count to the limit which is desirable.

Job timeout can be overridden in `.sync.yml`.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -4,6 +4,7 @@
   beaker_fact_matrix: {}
   excludes: []
   pidfile_workaround: false
+  timeout_minutes: 40
 .travis.yml:
   delete: true
 Gemfile:
--- a/moduleroot/.github/workflows/ci.yml.erb
+++ b/moduleroot/.github/workflows/ci.yml.erb
@@ -6,6 +6,7 @@ jobs:
   setup_matrix:
     name: 'Setup Test Matrix'
     runs-on: ubuntu-latest
+    timeout-minutes: <%= @configs['timeout_minutes'] %>
     outputs:
       beaker_setfiles: ${{ steps.get-outputs.outputs.beaker_setfiles }}
       puppet_major_versions: ${{ steps.get-outputs.outputs.puppet_major_versions }}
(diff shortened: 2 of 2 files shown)


traylenator:
I'm sure other folk have a better real world idea of what the default limit should be.

ekohl (changes requested):
I think tests often need more than 10 minutes. This feels way too aggressive to me and something we would need to modify on most repositories. I'd start with 30, 45 or 60 minutes by default.

I also wonder if we can query run times for our actions to make a more informed decision.

traylenator:
Start at 60 minutes and reduce by 10 each modulesync release till it becomes too painful.

I did have a quick look.

There's this 

https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#get-workflow-run-usage

bastelfreak:
https://github.com/voxpupuli/puppet-collectd/actions/runs/383544865 collectd for example took 16min to run and we've many more that take more than 10min.

traylenator:
```bash
curl \
  -H "Accept: application/vnd.github.v3+json" \
  https://api.github.com/repos/voxpupuli/puppet-nftables/actions/runs/1549477914/timing 
```

is not available, I think its only available for billable time in private repos.

traylenator:
> https://github.com/voxpupuli/puppet-collectd/actions/runs/383544865 collectd for example took 16min to run and we've many more that take more than 10min.

For that collected its the 12 minute Debian puppet 6 job (the slowest) that is significant rather than the total 16 minutes.

But sure push up the time.

bastelfreak (approved):

ekohl (approved):
I would be ok with 40 minutes as a first value.

PR #698, opened 2021-01-15
moduleroot/Dockerfile.erb: use ruby:2.7

kenyon:
This will use the current version of Ruby 2.7.

This change was inspired by the one originally done here: https://github.com/voxpupuli/puppet-puppetboard/pull/309

Also, closes #699.

### DIFF
--- a/moduleroot/Dockerfile.erb
+++ b/moduleroot/Dockerfile.erb
@@ -1,4 +1,4 @@
-FROM ruby:2.5.3
+FROM ruby:2.7
 
 WORKDIR /opt/puppet
 


moduleroot/Dockerfile.erb
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.5.3
+FROM ruby:2.5.8
```
ekohl, line 1: I wonder if it should just pull from `ruby:2.5` so we don't need to keep it updated manually:
```suggestion
FROM ruby:2.5
```

kenyon: Yes, good idea!

bastelfreak:
hey peoples.
The dockerfile contains `COPY . .` which copies the whole module root into the container. if you previously installed gems with `bundle install --path /something/withing/cwd` and are not on ruby 2.5, this will fail. I'm far away from being a Docker expert. Maybe there is a known workaround?

ekohl (approved):

ekohl:
>  The dockerfile contains `COPY . .` which copies the whole module root into the container. if you previously installed gems with `bundle install --path /something/withing/cwd` and are not on ruby 2.5, this will fail. I'm far away from being a Docker expert. Maybe there is a known workaround?

Hmm, not entirely sure. I personally don't use Docker here.

ekohl (approved):

dhollinger (approved):

PR #699, opened 2021-01-28
Install bundler2 for compatbility

dom-nie:
As this Dockerfile copies module root, if there is a lock file created with bundler2 docker will keep failing with 

> Step 9/11 : RUN bundle install
>  ---> Running in 15086922f6ce
> You must use Bundler 2 or greater with this lockfile.
> The command '/bin/sh -c bundle install' returned a non-zero code: 20

After this PR Dockerfile should be working with lock files generated with both bundler1 and bundler2

### DIFF
--- a/moduleroot/Dockerfile.erb
+++ b/moduleroot/Dockerfile.erb
@@ -1,4 +1,4 @@
-FROM ruby:2.5.3
+FROM ruby:2.7.2
 
 WORKDIR /opt/puppet
 


ekohl (commented):
How about we update it to Ruby 2.7 instead? That's what Puppet 7 uses and I think it includes Bundler 2 already.

dom-nie:
@ekohl That's right, Ruby 2.7 image has bundler 2.1.4 
I have changed the commit to upgrading the image to the latest version of 2.7

moduleroot/Dockerfile.erb
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.5.3
+FROM ruby:2.7.2
```
ekohl, line 1: https://github.com/voxpupuli/modulesync_config/pull/698 is open and drops the patch version. I'd prefer to do the same here. It also means this PR would supersede that PR (which is fine).

kenyon: I took care of it in #698 then :).

PR #706, opened 2021-03-23
changelog_generator: use upstream release

bastelfreak:
After many months of waiting, upstream releases 1.16.0 (and a day after
1.16.1). It looks like those releases contain all features we require
  and I didn't notice any regressions.


ekohl (approved):
Yay!

PR #709, opened 2021-05-25
add puppet-puppet_summary module

bastelfreak:


kenyon:
Reference: https://github.com/voxpupuli/puppet-puppet_summary

PR #710, opened 2021-06-13
allow all newer puppet gems

bastelfreak:


kenyon (approved):

ekohl (approved):
In Foreman we use `>=` instead of `~>` and I must say that I generally like it. in CI it's explicit anyway but as a developer I often just want the latest version. Something to consider.

config_defaults.yml
```diff
@@ -8,7 +8,7 @@
 .travis.yml:
   delete: true
 Gemfile:
-  puppet_version: '~> 6.0'
+  puppet_version: '>= 7.0'
```
ekohl, line 11: Should it be `>= 6.0` just to indicate we're still compatible with 6? It'll pull in 7 in most cases anyway.

bastelfreak: we can do that. in the past we updated it always to latest major version. but since it's now `>=` I think 6 is fine now.

PR #717, opened 2021-06-22
Move all static validations to the setup job

ekohl:
The default rake task is release_checks. This runs `lint`, `validate` prior to `parallel_spec` and ends with the `check` task. These static tasks shouldn't differ for the matrix job and can be run prior to running the whole test suite. Then in the actual test matrix, it only needs to run the tests (in parallel). This saves some CI resources when a static check fails.


bastelfreak (approved):

smortex (approved):
Makes sense!

PR #727, opened 2021-07-13
Add a job to warn about modulesync changes

ekohl:
01cfc20164bb4db5432b82d7a59e2227a718b240 added headers, but this isn't really a good experience for users. It's actually much better to fail in CI if a diff is detected. That gives them also a way to verify their changes to .sync.yml are correct.

This takes the approach of checking out the modulesync config with the version from .msync.yml and running a one off change in offline mode.  Then it uses git diff to see if there are differences. That should fail the build and also show the actual differences, which makes debugging easier.

Right now this is all theoretical and the principe should work, but isn't tested. I wanted to share it early on to open a discussion.

Another thing to think about is to make this easy for individual developers on their workstations. Something like a PDK mode for modulesync.


ghoneycutt:
Is there a way we can do this for one module and see what it looks like when you edit a file you shouldn't have?

moduleroot/.github/workflows/modulesync.yml.erb
```diff
@@ -0,0 +1,51 @@
...
+        uses: actions/checkout@v2
+        path: modules/${{ env.github_repository }}
+
+      - name: Setup ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.0'
+
+      - name: Set modulesync version
+        run: cat .msync.yml | ruby -ryaml -e 'puts "MSYNC_VER=#{YAML.load(STDIN)["modulesync_config_version"]}"' >> $GITHUB_ENV
```
bastelfreak, line 24: what about this:
```suggestion
        run: ruby -ryaml -e 'puts "MSYNC_VER=#{YAML.safe_load(File.read(".msync.yml"))["modulesync_config_version"]}"' >> $GITHUB_ENV
```

```console
$ ruby -ryaml -e 'puts "MSYNC_VER=#{YAML.safe_load(File.read(".msync.yml"))["modulesync_config_version"]}"'
MSYNC_VER=4.1.0
```

Allows us to spare one pipe

bastelfreak:
okay so this is really awesome!

ekohl:
> Is there a way we can do this for one module and see what it looks like when you edit a file you shouldn't have?

That was on my agenda: https://github.com/voxpupuli/puppet-example/pull/8. There it really shows this was 100% just a thought because the Github actions syntax is invalid :) I'll continue to iterate there.

ekohl:
I think I've worked out all the kinks in https://github.com/voxpupuli/puppet-example/pull/8 and this is now ready for review.

bastelfreak (approved):

moduleroot/.github/workflows/modulesync.yml.erb
```diff
@@ -0,0 +1,52 @@
...
+          ruby-version: '3.0'
+          bundler-cache: true
+          working-directory: msync_config
+
+      - name: Run modulesync
+        run: bundle exec msync update --offline --managed-modules-conf=pr-managed-modules.yaml --project-root=../modules
+        working-directory: msync_config
+
+      - name: Check for differences
+        run: git diff --exit-code
```
ekohl, line 51: Maybe we need to run `git add . && git diff --cached --exit-code` to also track added/removed files

moduleroot/.github/workflows/modulesync.yml.erb
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          path: modules/${{ github.repository }}
+
+      - name: Setup ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.0'
+
+      - name: Set modulesync version
+        run: ruby -ryaml -e 'puts "MSYNC_VER=#{YAML.safe_load(File.read(".msync.yml"))["modulesync_config_version"]}"' >> $GITHUB_ENV
```
ekohl, line 25: This would break with pre-release like in https://github.com/voxpupuli/puppet-telegraf/pull/174. Perhaps we need to update our `.msync.yml` file to write `git describe` there instead of hardcoding.

bastelfreak: I think the question is do we want to support prereleases, or do we want to be more strict with modulesync_config releases? I'm not sure here.

PR #729, opened 2021-07-24
Add script to adjust modules afer dep got migrated

bastelfreak:
the camptocamp/systemd module is now voxpupuli/systemd. I wrote a script
to update the namespace in .fixtures.yml, the namespace and upper
version in metadata.json and to create a pull request afterwards.

Call it like this:
```
export GITHUB_TOKEN=...
./bin/clean-git-checkouts
./bin/systemd_module_migration.sh
```

This script can be adjusted for other modules. None of our modules
install the systemd module in spec/spec_helper_acceptance.rb nor
spec/acceptance/*, so I didn't implement this.


bin/systemd_module_migration.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
smortex, line 1: Do we really want to use bash(1)?  If so:
```suggestion
#!/usr/bin/env bash
```

otherwise a fine:
```suggestion
#!/bin/sh
```

Most *BSD systems do not ship with bash(1) and do not install it in `/bin`, so I trend to always stick to using `/bin/sh` and avoid non-POSIX-compliant code :wink:

bin/systemd_module_migration.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# we rquire hub later on for the PR
+if ! which hub > /dev/null ; then
+  exit 1
+fi
```
smortex, line 6: Better tell the user what is wrong on stderr than in a comment :smile: 

```suggestion
if ! which hub > /dev/null ; then
  echo "We require hub later on for the PR" >&2
  exit 1
fi
```

bin/systemd_module_migration.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# identify all modules with systemd in .fixtures.yml
+while read -r file; do
+  sed -i 's#camptocamp/puppet-systemd#voxpupuli/puppet-systemd#' "${file}"
+done < <(grep -l "camptocamp/puppet-systemd" modules/voxpupuli/puppet-*/.fixtures.yml)
```
smortex, line 11: `<(...)` is a bashism.  If we stick to `/bin/sh` we end up with:

```sh
grep -l "camptocamp/puppet-systemd" modules/voxpupuli/puppet-*/.fixtures.yml | while read -r file ; do
```

But maybe we can just run sed(1) on all files, since files without match will not be modified?  Then the whole block become:

```sh
sed -i 's#camptocamp/puppet-systemd#voxpupuli/puppet-systemd#' modules/voxpupuli/puppet-*/.fixtures.yml
```

bin/systemd_module_migration.sh
```diff
@@ -0,0 +1,35 @@
...
+# commit and create a PR
+for module in modules/voxpupuli/puppet-* ; do
+  (cd "$module"
+  changed_file=$(git diff --name-only)
+  # last compare doesn't work
+  #if [ "${changed_file}" = ".fixtures.yml" ] || [ "${changed_file}" = "metadata.json" ] || [ "${changed_file}" = ".fixtures.yml metadata.json" ]; then
+  if [ "${changed_file}" != "" ]; then
+    git checkout -b systemd_voxpupuli origin/master
+    git commit -am 'switch from camptocamp/systemd to voxpupuli/systemd'
+    git push --set-upstream origin systemd_voxpupuli
```
ekohl, line 30: I tend to prefer avoid repeating the branch name
```suggestion
    git push --set-upstream origin HEAD
```

bin/systemd_module_migration.sh
```diff
@@ -0,0 +1,35 @@
...
+# bump the metadata because there was a recent major release
+./bin/bump-dependency-upper-bound puppet/systemd 4.0.0 modules/voxpupuli/puppet-*/metadata.json
+
+# commit and create a PR
+for module in modules/voxpupuli/puppet-* ; do
+  (cd "$module"
+  changed_file=$(git diff --name-only)
+  # last compare doesn't work
+  #if [ "${changed_file}" = ".fixtures.yml" ] || [ "${changed_file}" = "metadata.json" ] || [ "${changed_file}" = ".fixtures.yml metadata.json" ]; then
+  if [ "${changed_file}" != "" ]; then
```
ekohl, line 27: Does this work:
```suggestion
  git diff --exit-code .fixtures.yml metadata.json
  if [ $? != 0 ] ; then
```

bin/systemd_module_migration.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+if ! which hub > /dev/null ; then
+  echo "We require hub later on for the PR" >&2
+  exit 1
+fi
+
+# fix all modules with systemd in .fixtures.yml
+sed --in-place 's#camptocamp/puppet-systemd#voxpupuli/puppet-systemd#' modules/voxpupuli/puppet-*/.fixtures.yml
```
smortex, line 9: FreeBSD does not have long options… A more portable solution would be:

```suggestion
sed -I '' 's#camptocamp/puppet-systemd#voxpupuli/puppet-systemd#' modules/voxpupuli/puppet-*/.fixtures.yml
```

…but POSIX does not have in-place edit either (no `-I` / `-i`) so I am not sure about the best way to handle this if we want it to be portable.

Maybe just `set -e` at the beginning of the script to exit as soon as the first command fail?

ghoneycutt: You've got ruby involved with a puppet agent so you could use that :)

```
ruby -pi -e "gsub(/before/, 'after')" filename
```

ekohl:
So rather than building these one-off solutions I started https://github.com/ekohl/metadata_json_deps/pull/7. It would rely on the Puppet Forge to have them deprecated with the proper replacement, but it's scalable. I think the next step should be to introduce a "fix" mode in that gem so it can rectify the dependency. It could open the path to dependabot-style checks.

ekohl:
@bastelfreak thoughts on closing this? The check mode is currently there. We also migrated most if not all systemd modules. If we do have a new big change, I think it makes sense we focus our efforts on enhancing the generic replacement tool instead.

PR #731, opened 2021-07-27
Add a dummy test suite task to CI

ekohl:
This task is essentially a dummy but due to its dependencies it only completse if the entire test suite passed. This task can then be used as Github branch protection allowing the automerge feature to be used safely.

Tested in https://github.com/voxpupuli/puppet-example/pull/9

### DIFF
--- a/moduleroot/.github/workflows/ci.yml.erb
+++ b/moduleroot/.github/workflows/ci.yml.erb
@@ -100,3 +100,14 @@ jobs:
           BEAKER_FACTER_<%= fact.upcase %>: ${{ matrix.<%= fact %> }}
           <%- end -%>
 <%- end -%>
+
+  tests:
+    needs:
+      - unit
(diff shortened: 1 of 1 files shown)


bastelfreak:
I gave this a short try and github actions complained about the `run` statement, I think it requires a name. I updated it to this:
```
  tests:
    needs: unit, acceptance
    runs-on: ubuntu-latest
    name: Test suite
    steps:
      - name: dummy task
        run: echo Test suite completed
```

now I've:
```
The workflow is not valid. .github/workflows/ci.yml (Line: 80, Col: 12): Job 'tests' depends on unknown job 'unit, acceptance'.
```

ekohl:
Verified it against puppet-example and this is now ready.

ekohl:
See https://github.com/voxpupuli/puppet-example/actions/runs/1138783639 for a green run.

bastelfreak (approved):

Issue #733, opened 2021-07-29
puppet-strings prints warnings since the merge of #722


bastelfreak:
We had some discussions on the IRC and would like to do a new modulesync release. I notice that puppet-strings complains about comments since we merged https://github.com/voxpupuli/modulesync_config/pull/722

I'm not yet sure which file causes this
```
 $ /usr/bin/bundle exec rake reference
[warn]: Could not find file: DO
[warn]: Could not find file: NOT
[warn]: Could not find file: EDIT
[warn]: Could not find file: #
[warn]: Could not find file: https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
[warn]: The length of the summary for puppet_class 'prometheus::install' exceeds the recommended limit of 140 characters.
[warn]: The length of the summary for puppet_defined_type 'prometheus::scrape_job' exceeds the recommended limit of 140 characters.
[warn]: The length of the summary for puppet_class 'prometheus::unbound_exporter' exceeds the recommended limit of 140 characters.
[warn]: Invalid tag format for @example
[warn]: Invalid tag format for @example in file `manifests/blackbox_exporter.pp` near line 65
[warn]: Missing @param tag for parameter 'export_scrape_job' near manifests/blackbox_exporter.pp:65.
[warn]: Missing @param tag for parameter 'scrape_host' near manifests/blackbox_exporter.pp:65.
[warn]: Missing @param tag for parameter 'scrape_port' near manifests/blackbox_exporter.pp:65.
[warn]: Missing @param tag for parameter 'scrape_job_name' near manifests/blackbox_exporter.pp:65.
[warn]: Missing @param tag for parameter 'scrape_job_labels' near manifests/blackbox_exporter.pp:65.
[warn]: Missing documentation for Puppet type alias 'Prometheus::Uri' at types/uri.pp:1.
[warn]: Missing documentation for Puppet type alias 'Prometheus::GsUri' at types/gsuri.pp:1.
[warn]: Missing documentation for Puppet type alias 'Prometheus::S3Uri' at types/s3uri.pp:1.
[warn]: Missing documentation for Puppet type alias 'Prometheus::Initstyle' at types/initstyle.pp:1.
Files:                   52
Modules:                  0 (    0 undocumented)
Classes:                  0 (    0 undocumented)
Constants:                0 (    0 undocumented)
Attributes:               0 (    0 undocumented)
Methods:                  0 (    0 undocumented)
Puppet Classes:          44 (    0 undocumented)
[...]

PR #734, opened 2021-07-29
Delete legacy .yardopts

bastelfreak:
I think we don't need thefile at the moment. We don't publish html docs
anymore, only the REFERENCE.md. `bundle exec rake reference` does not
require the .yardopts file. I tested this at https://github.com/voxpupuli/puppet-lldpd/pull/106

Fixes https://github.com/voxpupuli/modulesync_config/issues/733

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -98,5 +98,7 @@ spec/spec_helper_acceptance.rb:
   unmanaged: true
 CONTRIBUTING.md:
   delete: true
+.yardopts:
+  delete: true
 ...
 # vim: syntax=yaml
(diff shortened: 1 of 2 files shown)


ekohl (approved):

ekohl:
Tracing it back, it was added in 9e927f44f26febe79c5840170b8d7394cda5ad33 and then extended in 7139ff43b758d4589671bc80ff714fc7e7952727 but by now we've decided that generating HTML docs in the repo is a bad idea. I'm also not sure why the markup option would be needed. We don't have it in Foreman and I never noticed it. I guess it may have been needed in the past.

PR #745, opened 2021-08-17
Limit github_changelog_generator to Ruby 2.5+

ekohl:
This keeps it possible to test on Puppet 5.


bastelfreak (approved):

ekohl:
Verified in https://github.com/voxpupuli/puppet-example/actions/runs/1138783639

PR #758, opened 2021-11-02
implement modulespecific puppet-lint plugins

bastelfreak:


bastelfreak:
I think voxpupuli-puppet-lint-plugins should contain all recommended plugins / all plugins that help us to enforce our styleguide from https://voxpupuli.org/docs/reviewing_pr/. Some modules require some  rework before all linter plugins pass. To not block their CI pipeline, I would like to enable us to make them opt-in. This patch classifies disabled plugins. Modules can enable them individually. When all modules are updated we can remove the plugins from the blocklist. I think this helps us to streamline our setup and enabled voxpupuli-puppet-lint-plugins to bundle all plugins we recommend, not only the core plugins.

moduleroot/Rakefile.erb
```diff
@@ -69,4 +69,23 @@ begin
...
+  parameter_types
+]
+
+# A list of puppet-lint plugins we enable for this specific module
+modulespecific_puppet_lint_checks = %w[
+<%- [@configs['puppet_lint_checks']].flatten.compact.each do |line| -%>
+  <%= line %>
+<%- end -%>
+]
+puppet_lint_disable_checks = voxpupuli_puppet_lint_disable_checks - modulespecific_puppet_lint_checks
```
ekohl, line 87: Can't we do this in ERB and only output the final configuration?

ekohl (changes requested):
While I normally prefer fewer files, in this case I'm wondering about a `.puppet-lint.rc` file. In Foreman's modulesync we have this: https://github.com/theforeman/foreman-installer-modulesync/blob/master/moduleroot/.puppet-lint.rc.erb

To me the benefit is that if you have other tooling which calls puppet-lint (like a vim plugin) it also respects the configuration.

bastelfreak:
while implementing I already wondered why we don't have a .puppet-lint.rc (anymore). I will update it to include one. That also enables us to add `--fail-on-warnings` which might be better placed here and not in https://github.com/voxpupuli/voxpupuli-test/pull/59

moduleroot/Rakefile.erb
bastelfreak: updated it

bastelfreak:
I would like to add `--fail-on-warnings` for a cleaner changelog later on, in another PR.

moduleroot/.puppet-lint.rc.erb
```diff
@@ -0,0 +1,4 @@
+<% checks = @configs['disabled_lint_checks'] - ( @configs['enabled_lint_checks'] || [] ) -%>
```
ekohl, line 1: Not strictly against, Is this needed? A module can also override the `disabled_lint_checks` and include the defaults then. Perhaps that keeps it simple.

If you do keep this, I wonder about the discoverability. Perhaps it should be
```suggestion
<% checks = @configs['disabled_lint_checks'] - @configs['enabled_lint_checks'] -%>
```

Then have a default empty array in `config_defaults.yml`.

bastelfreak: I think having the disabled list not overwritten by individual modules is more future proof for new voxpupuli-puppet-lint-releases.

ekohl (approved):
I do wonder about the path to get these enabled on all modules. Would a blog post about this help?

bastelfreak:
I can contribute a blogpost after we released this + a new voxpupuli-puppet-lint-plugins version.

ekohl:
That would be great.

sebastianrakel (approved):

PR #763, opened 2021-11-15
Switch to github.ref_name for concurrency group

root-expert:
`github.head_ref` exists only on Pull Requests. This can cause problems
e.g if we want to run the CI nightly on a cron job. `github.ref_name`
seems to be the same thing.

### DIFF
--- a/moduleroot/.github/workflows/ci.yml.erb
+++ b/moduleroot/.github/workflows/ci.yml.erb
@@ -7,7 +7,7 @@ name: CI
 on: pull_request
 
 concurrency:
-  group: ${{ github.head_ref }}
+  group: ${{ github.ref_name }}
   cancel-in-progress: true
 
 jobs:


bastelfreak:
can you please test this on puppet-example?

root-expert:
Works as expected for PRs. [Previous](https://github.com/voxpupuli/puppet-example/actions/runs/1461894510) run was cancelled by the [newer](https://github.com/voxpupuli/puppet-example/actions/runs/1461895497).

Maybe we should test it for nightly as well

root-expert:
Opened https://github.com/voxpupuli/puppet-example/pull/17

ekohl (approved):

root-expert:
This change was proposed in case cron schedule was added to our current CI pipeline.
If that doesn't happen eventually, are current setup should work as expected as long as we are running the pipeline only on PRs.

root-expert:
Maybe we should include this to gha-puppet repo since this is going to be deprecated?

ekohl:
Actually, it'll remain relevant. This needs to be specified in the workflow itself, not in the shared workflow. You may see that https://github.com/voxpupuli/modulesync_config/pull/771 does not modify the section so I'm going to merge this. I just missed that you marked it as ready for review.

PR #780, opened 2022-05-16
Add additional_packages config option

ekohl:
This was added to gha-puppet and should be supported.


bastelfreak (approved):
tested this on https://github.com/voxpupuli/puppet-augeasproviders_apache/pull/11, works fine.

PR #788, opened 2022-07-22
Update REFERENCE.md during CHANGELOG.md creation

bastelfreak:
At the moment our release rake tasks checks if the REFERENCE.md is
up2date. It doesn't update it. I think it makes sense to generate the
REFERENCE.md in the same step as the CHANGELOG.md, because that will be
reviewed in a PR. This commit implements it. I created
https://github.com/voxpupuli/puppet-logrotate/pull/201 to check if
rubocop is happy.

### DIFF
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -55,6 +55,11 @@ begin
     config.project = metadata.metadata['name']
   end
 
+  # update the REFERENCE.md if it exists
+  # reference task will run before changelog
+  # https://dev.to/molly/rake-task-enhance-method-explained-3bo0
+  Rake::Task['changelog'].enhance(['reference']) if File.exist?('REFERENCE.md')
+
   # Workaround for https://github.com/github-changelog-generator/github-changelog-generator/issues/715
   require 'rbconfig'
   if RbConfig::CONFIG['host_os'] =~ /linux/
@@ -66,7 +71,6 @@ begin
       File.open(changelog_file, "w") {|file| file.puts new_contents }
     end
   end
-
 rescue LoadError
 end
 # vim: syntax=ruby


smortex (commented):
Looks hackish but it would make my life easier avoiding me to re-program my brain :+1:  :trollface:.

I have mixed feelings because this makes two supposedly independent things managed by a single task whose name suggest it only takes care of one of them.  Worse, as I understand it, if the "sidecar" task fail, the changelog is not generated.  I am quite concerned about this.

I suppose that a more "elegant" way (less hackish) would be to rework task names in some way like:

* Rename `changelog` to `porcelain::changelog`
* Rename `reference` to `porcelain::reference`
* Add `prerelease` that run both `porcelain::reference` and `porcelain::changelog`
* Add `reference` that says it is deprecated and to use `prerelease` of `porcelain::reference`
* Add `changelog` that says it is deprecated and to use `prerelease` of `porcelain::changelog`

Much more work and maybe not worth it.  For voxpupuli maintainers, that would mean to use `prerelease` when creating a new release PR (instead of `changelog` + `reference` currently) and `release` after it is merged.  The day we need to add another bunch of actions, we can extend the existing interface and add more "porcelain" commands.

What people think about this?  If we choose to go in this direction, I am of course willing to take care of it.

ekohl:
As a principle I like combining these, but I don't like that we have a rather large amount of modulesynced code here. Is it perhaps time for a `voxpupuli-development` or push this into https://github.com/voxpupuli/voxpupuli-release? A task `release:prepare` would also make sense to me. It does mean the release task needs to pull in many more deps, which I don't like as much.

bastelfreak:
I'm also not the biggest fan of extending the `changelog` rake task. But it has the benefit that people can keep using their usual workflow. We could also add a `release::prepare` job and raise a warning in `changelog` like `this task is deprecated, use release::prepare`?

ekohl:
Just to be explicit: IMHO we should also move the changelog definition into a gem. It's a rather large piece of code that's always the same for us.

smortex:
I started to work on it:
* https://github.com/voxpupuli/voxpupuli-release/pull/44
* https://github.com/voxpupuli/modulesync_config/pull/789

bastelfreak:
I think this is now obsolete

PR #791, opened 2022-08-17
anchor most .pmtignore patterns to the root of the module

jhoblitt:
The PathSpec used by .pmtignore will match all recursive directories
unless there is a "separator" at the start of the pattern per
https://git-scm.com/docs/gitignore.

As an example of the problem, the pattern `log/` will match a dir named
`foo/log`. Prefixing the pattern with `/` forces the pattern to match
only relative to the root of the module. Where as the pattern `/log/`
would not match a dir named `foo/log`.

### DIFF
--- a/moduleroot/.pmtignore.erb
+++ b/moduleroot/.pmtignore.erb
@@ -1,40 +1,40 @@
 # Managed by modulesync - DO NOT EDIT
 # https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
 
-docs/
-pkg/
-Gemfile
-Gemfile.lock
-Gemfile.local
-vendor/
-.vendor/
-spec/
-Rakefile
-.vagrant/
-.bundle/
-.ruby-version
-coverage/
-log/
-.idea/
-.dependencies/
-.github/
-.librarian/
-Puppetfile.lock
+/docs/
+/pkg/
+/Gemfile
+/Gemfile.lock
+/Gemfile.local
+/vendor/
+/.vendor/
+/spec/
+/Rakefile
+/.vagrant/
+/.bundle/
+/.ruby-version
+/coverage/
+/log/
+/.idea/
+/.dependencies/
+/.github/
+/.librarian/
+/Puppetfile.lock
 *.iml
-.editorconfig
-.fixtures.yml
-.gitignore
-.msync.yml
-.overcommit.yml
-.pmtignore
-.rspec
-.rspec_parallel
-.rubocop.yml
-.sync.yml
+/.editorconfig
+/.fixtures.yml
+/.gitignore
+/.msync.yml
+/.overcommit.yml
+/.pmtignore
+/.rspec
+/.rspec_parallel
+/.rubocop.yml
+/.sync.yml
 .*.sw?
-.yardoc/
-.yardopts
-Dockerfile
+/.yardoc/
+/.yardopts
+/Dockerfile
 <% if ! @configs['paths'].nil? -%>
 <% @configs['paths'].each do |path| -%>
 <%= path %>


bastelfreak:
I think this makes sense. Did you compare it to a .pmtignore that pdk creates?

moduleroot/.pmtignore.erb
```diff
@@ -1,39 +1,39 @@
...
+/.rspec
+/.rspec_parallel
+/.rubocop.yml
+/.sync.yml
 .*.sw?
-.yardoc/
-.yardopts
+/.yardoc/
+/.yardopts
 Dockerfile
```
ekohl, line 37: Any reason this isn't related to the root?

ekohl (approved):
I'd be OK with this. The puppet-modulebuilder gem (which is actually used nowadays) also has files including `/`:
https://github.com/puppetlabs/puppet-modulebuilder/blob/9a1edcaf576ee2a98d64ed23de4caad06a07ff72/lib/puppet/modulebuilder/builder.rb#L8-L17 

I also wonder if we should also ignore `/.*` like the default. If we do that, I wonder how many files in this list remain and how many of those we can push upstream.

For example, `.*.sw?` are vim swap files. Nobody wants those. Perhaps we can even introduce some option to remove all development related ones (`Gemfile*`, `spec`). In other words, get rid of `.pmtignore` and push it into our tooling.

jhoblitt:
> I think this makes sense. Did you compare it to a .pmtignore that pdk creates?

I have not but doing so makes sense before merging this.

jhoblitt:
> I also wonder if we should also ignore `/.*` like the default. If we do that, I wonder how many files in this list remain and how many of those we can push upstream.

What about going a step further and ignoring everything by default and then negating only a few select filetypes/paths? E.g. `!/manifests/**/*.pp`, `!/lib/**/*.rb`, `!/files/**`, `!/templates/**/*.erb`, `!/templates/**/*.epp`, etc.

bastelfreak:
I like the idea of blocking everything and working with an explicit allowlist. do you know how the forge handles it? can I upload modules with a good old zipbomb?

ekohl:
> What about going a step further and ignoring everything by default and then negating only a few select filetypes/paths?

Does that actually work?

jhoblitt:
Obviously, we need to test it but if it doesn't work then I suspect that would be considered a bug in https://rubygems.org/gems/pathspec.

jhoblitt:
https://github.com/highb/pathspec-ruby/blob/2aa17707e4c029da35c6ae974eed265f2a8b4ebf/lib/pathspec/gitignorespec.rb#L38-L42

jhoblitt:
Related to #809

bastelfreak:
@jhoblitt I think this idea is good and we should move forward with it. In addition to your suggested include list we need to add the LICENSE file (some licenses require us to ship the actual license with the product). I don't know how the forge handles puppet-strings. the REFERENCE.md probably needs to be included as well.

moduleroot/.pmtignore.erb
jhoblitt: No... fixed and repushed.

jhoblitt:
Ugh, the CI failure is repeatable locally:

```bash
$ bundle exec msync update --noop --git-base=https://github.com/
Syncing 'puppet-allknowingdns'
bundler: failed to load command: msync (/home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/bin/msync)
NoMethodError: undefined method `[]' for nil:NilClass
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/lib/modulesync/repository.rb:37:in `default_branch'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/lib/modulesync/repository.rb:55:in `switch'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/lib/modulesync/repository.rb:86:in `prepare_workspace'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/lib/modulesync.rb:122:in `manage_module'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/lib/modulesync.rb:178:in `block in update'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/lib/modulesync.rb:177:in `each'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/lib/modulesync.rb:177:in `update'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/lib/modulesync/cli.rb:149:in `update'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/thor-1.2.1/lib/thor/command.rb:27:in `run'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/thor-1.2.1/lib/thor/invocation.rb:127:in `invoke_command'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/thor-1.2.1/lib/thor.rb:392:in `dispatch'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/thor-1.2.1/lib/thor/base.rb:485:in `start'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/lib/modulesync/cli/thor.rb:21:in `start'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/gems/modulesync-2.5.0/bin/msync:24:in `<top (required)>'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/bin/msync:23:in `load'
  /home/jhoblitt/github/modulesync_config/.bundle/ruby/2.7.0/bin/msync:23:in `<top (required)>'
```

jhoblitt:
Changes to the `git` gem seem to be at fault. Pinning the version to `1.13.0` fixes it.  The breakage appears to be `> 1.13.0` and `<  1.17.0`.

bastelfreak (approved):

bastelfreak:
I think this is a good way forward. In the future we should check to reverse it and block everything that isn't explicitly allowed (and I wand the same in r10k and code-manager, code-manager already did some similar changes in the beginning of 2022).

PR #796, opened 2022-09-12
add Rakefile config.project to set changelog project name

jhoblitt:
For example, in `.sync.yml`:

    Rakefile:
      config.project: puppet-ipmi

### DIFF
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -51,7 +51,7 @@ begin
     config.future_release = "v#{metadata.version}" if metadata.version =~ /^\d+\.\d+.\d+$/
     config.header = "# Changelog\n\nAll notable changes to this project will be documented in this file.\nEach new release typically also includes the latest modulesync defaults.\nThese should not affect the functionality of the module."
     config.exclude_labels = %w{duplicate question invalid wontfix wont-fix modulesync skip-changelog}
-    config.user = '<%= @configs['config.user'] -%>'
+    config.user = '<%= @configs['config.user'] || @configs[:namespace] %>'
     config.project = '<%= @configs['config.project'] || @configs[:puppet_module] %>'
   end
 
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -52,7 +52,7 @@ begin
     config.header = "# Changelog\n\nAll notable changes to this project will be documented in this file.\nEach new release typically also includes the latest modulesync defaults.\nThese should not affect the functionality of the module."
     config.exclude_labels = %w{duplicate question invalid wontfix wont-fix modulesync skip-changelog}
     config.user = '<%= @configs['config.user'] -%>'
-    config.project = metadata.metadata['name']
+    config.project = '<%= @configs['config.project'] || @configs[:puppet_module] %>'
   end
 
   # Workaround for https://github.com/github-changelog-generator/github-changelog-generator/issues/715
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -35,7 +35,6 @@ Gemfile:
       - gem: puppet-strings
         version: '>= 2.2'
 Rakefile:
-  config.user: voxpupuli
   puppet_strings_patterns: []
 .puppet-lint.rc:
   disabled_lint_checks:


moduleroot/Rakefile.erb
```diff
@@ -52,7 +52,7 @@ begin
     config.header = "# Changelog\n\nAll notable changes to this project will be documented in this file.\nEach new release typically also includes the latest modulesync defaults.\nThese should not affect the functionality of the module."
     config.exclude_labels = %w{duplicate question invalid wontfix wont-fix modulesync skip-changelog}
     config.user = '<%= @configs['config.user'] -%>'
-    config.project = metadata.metadata['name']
+    config.project = <%= @configs['config.project'] || "metadata.metadata['name']" %>
```
smortex, line 55: Some extra quoting is needed I think:
```suggestion
    config.project = <%= @configs['config.project']&.inspect || "metadata.metadata['name']" %>
```

ekohl: I think it's safe to assume a project doesn't have `'` in it, so this would also work:
```suggestion
    config.project = '<%= @configs['config.project'] || metadata.metadata['name'] %>'
```

jhoblitt: I like both of the proposed approaches.  I think @ekohl's approach might be slightly better in that in results in single quotes, which would be better for if/when we start allowing `rubocop` to lint the `Rakefile`.

jhoblitt:
:facepalm: `metadata.metadata['name']` needs to be a literal string...

jhoblitt:
3rd time is the charm...

jhoblitt:
Nope... because `!!''` is true.

jhoblitt:
I've manually tested the 4th attempt as working both when `config.project` is set and it is not.  This is probably a sign that we need a test harness in this repo...

moduleroot/Rakefile.erb
```diff
@@ -52,7 +52,7 @@ begin
     config.header = "# Changelog\n\nAll notable changes to this project will be documented in this file.\nEach new release typically also includes the latest modulesync defaults.\nThese should not affect the functionality of the module."
     config.exclude_labels = %w{duplicate question invalid wontfix wont-fix modulesync skip-changelog}
     config.user = '<%= @configs['config.user'] -%>'
-    config.project = metadata.metadata['name']
+    config.project = <%= @configs['config.project'] ? "'#{@configs['config.project']}'" : "metadata.metadata['name']" %>
```
ekohl, line 55: Modulesync also has this info. Twice actually:

```suggestion
    config.project = '<%= @metadata[:module_name] %>'
```

That's defined [here](https://github.com/voxpupuli/modulesync/blob/ac051ca961c1119a4b731bf990e8db0be6cf497b/lib/modulesync.rb#L95-L100)

It's also in `@configs` which would be more consistent with the previous line:
```suggestion
    config.project = '<%= @configs[:puppet_module] %>'
```

IMHO metadata is cleaner, but we should also add namespace to `metadata`: https://github.com/voxpupuli/modulesync/pull/254

moduleroot/Rakefile.erb
```diff
@@ -52,7 +52,7 @@ begin
     config.header = "# Changelog\n\nAll notable changes to this project will be documented in this file.\nEach new release typically also includes the latest modulesync defaults.\nThese should not affect the functionality of the module."
     config.exclude_labels = %w{duplicate question invalid wontfix wont-fix modulesync skip-changelog}
     config.user = '<%= @configs['config.user'] -%>'
```
ekohl, line 54: I realized we can remove our hardcoding in `config_defaults.yml` for this:
```suggestion
    config.user = '<%= @configs['config.user'] || @configs[:namespace] %>'
```

ekohl (commented):
I realized this is another way to go: rely on native modulesync which already has the config.

moduleroot/Rakefile.erb
jhoblitt: I really like this. This should work for non-voxpupuli users without having to put anyting in the `.sync.yml`. :100:

jhoblitt: I'm going to change this slightly so that it can be overridden, if need be.

jhoblitt:
@ekohl I've implemented your changes and removed `config.user` from `config_defaults.yml`. The diff from `--noop` looks good to me:

```Syncing 'puppet-allknowingdns'
Using no-op. Files in 'puppet-allknowingdns' may be changed but will not be committed.
Files changed:
diff --git a/Rakefile b/Rakefile
index f92f051..e5863a2 100644
--- a/Rakefile
+++ b/Rakefile
@@ -52,7 +52,7 @@ begin
     config.header = "# Changelog\n\nAll notable changes to this project will be documented in this file.\nEach new release typically also includes the latest modulesync defaults.\nThese should not affect the functionality of the module."
     config.exclude_labels = %w{duplicate question invalid wontfix wont-fix modulesync skip-changelog}
     config.user = 'voxpupuli'
-    config.project = metadata.metadata['name']
+    config.project = 'puppet-allknowingdns'
   end
 
   # Workaround for https://github.com/github-changelog-generator/github-changelog-generator/issues/715
Files added:


--------------------------------
```

ekohl (approved):

ekohl:
It does want signed commits in this repo.

PR #798, opened 2022-10-03
Allow opting-out of GitHub Action acceptance tests

smortex:
Some modules acceptance tests cannot be run from a docker container as exposed by GitHub Actions (e.g. the puppet-virtualbox module want to install the running kernel headers to build kernel modules (which is not possible if the kernel of the host system is not one of the kernels available in the test system), and then load modules in the kernel which is not allowed.

Add a configuration flag `acceptance_tests` (defaulting to true) and make modulesync check it before emitting the suitable GitHub Actions for the module.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -4,6 +4,7 @@
   excludes: []
   pidfile_workaround: false
   additional_packages: ''
+  acceptance_tests: true
 .travis.yml:
   delete: true
 Jenkinsfile:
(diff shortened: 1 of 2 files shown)


config_defaults.yml
```diff
@@ -4,6 +4,7 @@
   excludes: []
   pidfile_workaround: false
   additional_packages: ''
+  enable_acceptance_tests: true
```
ekohl, line 7: I'd consider
```suggestion
  acceptance_tests: true
```

ekohl (approved):

PR #800, opened 2022-11-02
puppet-strings: require 4.x

bastelfreak:
puppet-strings 3 changed how links in markdown are represented. To not cause a lot of noise with puppet-strings 2 vs 3, we should enforce we're on puppet-strings 3.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -36,6 +36,8 @@ Gemfile:
         version: '~> 2.0'
       - gem: faraday-retry
         version: '~> 2.1'
+      - gem: puppet-strings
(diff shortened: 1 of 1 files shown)


ekohl (commented):
Note that 3.0 requires Puppet 2.7. I think we exclude the release group on Puppet 6 tests, but I don't know for sure if it properly resolves. Perhaps that's worth a try on https://github.com/voxpupuli/puppet-example?

bastelfreak:
meh it fails ruby 2.5: https://github.com/voxpupuli/puppet-example/pull/24

sebastianrakel (approved):

PR #806, opened 2023-01-06
Delete legacy SECURITY.md

bastelfreak:
We now have that on our org level:

https://github.com/voxpupuli/.github/pull/3


ekohl (approved):
Yay for fewer files.

PR #807, opened 2023-01-10
Add puppet-format

ekohl:
Previously this was managed via [pdksync_config](https://github.com/voxpupuli/pdksync_config), but it's literally the only module managed that way. By using modulesync we standardize on a single layout.

I've also sorted the managed modules.


ekohl:
It actually already was modulesynced once. https://github.com/voxpupuli/puppet-format/pull/11 updates it to the latest version.

tuxmea (approved):

PR #816, opened 2023-03-14
Add notes on using dev containers

NITEMAN:


moduleroot/.github/CONTRIBUTING.md.erb
```diff
@@ -249,3 +250,28 @@ For more information and tips & tricks, see [voxpupuli-acceptance's documentatio
...
+  -e DOCKER_IN_DOCKER=true \
+  puppet/puppet-dev-tools:4.x \
+    /bin/bash -c " \
+      git config --global --add safe.directory /repo; \
+      bundle config set --local path '.vendor/'; \
+      bundle config set --local without 'development release'; \
+      bundle config set --local with 'system_tests': \
+      bundle install --jobs "$(nproc)"; \
+      bundle update; \
+      bundle clean ; \
```
ekohl, line 273: Why is it needed to run bundle install, update and clean? If it's a fresh install that should be redundant, right?

ekohl (commented):
This suggests we should drop the `Dockerfile` we now sync to modules since it's redundant.

moduleroot/.github/CONTRIBUTING.md.erb
NITEMAN: I think quite not, yesterday I was bitten by lock file not including puppet strings 3

On IRC @bastelfreak told me:

> "it's complcated"
> a) we don't ship a Gemfile.lock in the modules ,that would cause more trouble
> we cannot enforce puppet-strings 3 everywhere because that requires Ruby 2.7 and that will break Puppet 6 tests or people using ancient operating systems like... RHEL8

So I played for safe and ported your "all in one solution if you don't know if you need to install or update gems:": https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.github/CONTRIBUTING.md.erb#L143

NITEMAN:
Regarding the Dockerfile, it doesn't work right now (at least for elasticsearch module). Also tried that yesterday:
> #0 5.641 You have requested:
> #0 5.641   puppet ~> 6.0
> #0 5.641
> #0 5.641 The bundle currently has puppet locked at 7.23.0.
> #0 5.641 Try running `bundle update puppet`

moduleroot/.github/CONTRIBUTING.md.erb
ekohl: Oh yes, it mounts the repo as a volume so that can contain `Gemfile.lock`. I'll admit I really dislike containers for development work since I have my environment set up just the way I like it and containers never have that. I'm certainly not the target audience of this.

NITEMAN: Certainly, but for someone like me... which doesn't have the time/expertise to set up a proper environment but wants to contribute some work... containers are a silver bullet to get up&running in almost no time

NITEMAN: Another option would be publishing a voxpupuli container for quick development

ekohl:
In https://github.com/voxpupuli/modulesync_config/pull/827 the Dockerfile was removed.

PR #823, opened 2023-04-18
Add faraday-retry to release gem group

bastelfreak:
This is an optional requirement for the changelog generator. GCG throws a warning when the gem is missing. Test PR: https://github.com/voxpupuli/puppet-example/pull/31

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -36,6 +36,10 @@ Gemfile:
         ruby-operator: '>='
       - gem: voxpupuli-release
         version: '~> 2.0'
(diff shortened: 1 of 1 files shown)


ekohl (approved):
I'm not happy about it, but I think it's the least of 2 evils now. Might be good to make GHCG depend on faraday-retry, but this could be merged as a stop gap solution.

bastelfreak:
@chelnak worked on a replacement for GCG and I think Puppet already uses it. Maybe that's a good long term solution.

PR #824, opened 2023-04-27
Gemfile: Pin to Puppet ~> 7.24

bastelfreak:
This is the first step to deprecate puppet 6. We raise the default puppet version in the Gemfile in our modules. also we limit it to 7.x so we don't pull in puppet 8 by accident. And we pin to 7.24 or newer to prevent issues with ruby-concurrent.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -13,7 +13,7 @@
 Jenkinsfile:
   delete: true
 Gemfile:
-  puppet_version: '>= 6.0'
+  puppet_version: '~> 7.24'
   required:
     ':test':
       - gem: voxpupuli-test


smortex (approved):
Looks good. I guess we will change this back to a `>=` at some point when Puppet 8 has settled?

bastelfreak:
I'm not sure yet. We saw breaking CI pipelines because we had `>= 6` and unexpectedly got puppet 8. In gems we try to set upper versions. I think `~> 7.24` is fine now and as soon as puppet 8 is available we could move to `~> 8.0.1`. This is just a default/fallback. It can be overwritten via environment variable (and we do that in CI).

config_defaults.yml
```diff
@@ -13,7 +13,7 @@
 Jenkinsfile:
   delete: true
 Gemfile:
-  puppet_version: '>= 6.0'
+  puppet_version: '~> 7.24'
```
ekohl, line 16: Previously it also allowed Puppet 8, now it no longer does. Is this intended?

bastelfreak: indeed. we allowed puppet 8 and that break our CI. To prevent that in the future it makes sense to limit to a single version. and the gemfile has an environment variable to overwrite it.

PR #836, opened 2023-06-19
CI: Run metadata_deps on all modules

bastelfreak:


ekohl (commented):
Bit late to the party, but I feel like this should be its own workflow. They don't really depend on each other, do they?

PR #837, opened 2023-06-22
add modules param to spec_helper_acceptance.rb.erb

jhoblitt:


jhoblitt:
Another option would be a flag to remove `configure_beaker` from `spec_helper_acceptance.rb` completely.

moduleroot/spec/spec_helper_acceptance.rb.erb
```diff
@@ -5,6 +5,10 @@
 
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
+<%- unless @configs['modules'] -%>
 configure_beaker
+<%- else -%>
+configure_beaker(modules: <%= @configs['modules'] %>)
```
bastelfreak, line 11: what about always setting @configs['modules'] here and add a default value to config_defaults.yaml? that reduces the logic here.

jhoblitt: I'm fine with that. I did it the way it is in the interest of not causing a diff on the next msync run for no functional change.

jhoblitt:
Take #2 now with the ability to remove `#configure_breaker` completely. I've discovered I need this ability to use `#scp_to`.

jhoblitt:
Grrr, rubocop is unhappy about the extra newline.  Standby for take #3.

bastelfreak (approved):

moduleroot/spec/spec_helper_acceptance.rb.erb
```diff
@@ -5,6 +5,8 @@
 
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
-configure_beaker
+<%- unless @configs['configure_beaker'].nil? -%>
```
bastelfreak, line 8: thinking about it again... if it's set to `:fixtures` we could automatically add `task :beaker => "spec_prep"` to the rakefile? But that would then require puppetlabs_spec_helper :thinking:

jhoblitt: Yes, the gem would need to be added.  Here's a PR I'm currently working on that had `.sync.yml` changes (rewritten by psych) and adds `./rakelib/beaker.rake`. https://github.com/lsst-it/puppet-ccs_software/pull/39/files

jhoblitt: That PR is also a good example of the pain of `.fixtures.yml` not having any dep resolution.  I don't understand why `Puppetfile` never became a thing for test fixtures...

ekohl: Probably because r10k doesn't do dependency resolution either. Does `Puppetfile` support symlinks?

ekohl: > thinking about it again... if it's set to `:fixtures` we could automatically add `task :beaker => "spec_prep"` to the rakefile? But that would then require puppetlabs_spec_helper thinking

Perhaps this bit would be useful in `Rakefile`: https://github.com/theforeman/foreman-installer-modulesync/blob/5bff8e9a2800ea6e0928aaaa0c2778c60d426bc1/moduleroot/Rakefile.erb#L21-L22

PR #838, opened 2023-06-22
fix yamllint "truthy" warnings

jhoblitt:


moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -4,7 +4,7 @@
 
 name: CI
 
-on: pull_request
```
bastelfreak, line 7: did you see errors about this? I haven't seen quotes keys anywhere else.

jhoblitt: Yes, the yamllint action I use warns about it: 

image

jhoblitt:
```---
# yamllint disable rule:quoted-strings
name: yamllint

"on":
  - push

jobs:
  yamllint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2

      - name: Run yamllint
        uses: bewuethr/yamllint-action@v1

Comment (smortex) [reply moduleroot/.github/workflows/ci.yml.erb:7]: lol, YAML is so :clown_face: 

```
irb(main):005:0> YAML.load('on: no')
=> {true=>false}
irb(main):006:0> YAML.load('"on": no')
=> {"on"=>false}
```

I guess this PR makes sense, but I also have never saw this into the wild at GitHub.  Could you test it?

Comment (ekohl) [reply moduleroot/.github/workflows/ci.yml.erb:7]: Yes, YAML is crazy. I guess GH did some special casing for `on`.

Comment (jhoblitt) [reply moduleroot/.github/workflows/ci.yml.erb:7]: It seems like github must have put a special case in their parser. Regardless, "on" always causes an error to show in my text editor so I've started quoting it in all my gha workflows.

Comment (wyardley) [reply moduleroot/.github/workflows/ci.yml.erb:7]: I have seen this error in yamllint a number of times. it's totally harmless (having `on` there) in the context of GitHub Actions configs, but annoying. Luckily, it's a warning vs. a fatal error in the default configuration, so I normally just ignore it. but adding an explicit ignore doesn't seem like a bad idea to me.

This is another thing that's legal in GHA configs, but kind of ugly:
```yaml
on:
  push:
    branches:
    - main
  pull_request:
```

Since in this case, you can't do just `on: pull_request` in the examples above because there are multiple values.

I have found it is legal to do
```yaml
on:
  pull_request: {}
```
though, which can work as a workaround for it.

Another random aside for people who really like checking stuff as far left as possible: if you use `pre-commit`, you can add this to your config to do some additional validation of the GHA configs

```yaml
  - repo: https://github.com/python-jsonschema/check-jsonschema
    rev: 0.28.5
    hooks:
      - id: check-github-workflows
```

wyardley: side note: would it make sense to just silence the truthy values warning vs. quoting `"on"` for that line specifically? I personally find the quoted `"on"` a bit weird - I can throw up a PR so people can see if they like that better?

wyardley: #927 is how it would look with the ignore. This one is probably more "correct", and the quoted key doesn't seem to introduce a new error, so I'm good with either way personally.

wyardley (approved):

wyardley:
I think closing this is ok since #927 went in?

PR #840, opened 2023-06-30
CI: Make runners configureable

bastelfreak:
Tested in https://github.com/voxpupuli/puppet-nginx/pull/1562

### DIFF
--- a/moduleroot/.github/workflows/ci.yml.erb
+++ b/moduleroot/.github/workflows/ci.yml.erb
@@ -35,6 +35,6 @@ jobs:
 <%- if @configs.key?('rubocop') -%>
       rubocop: <%= @configs['rubocop'] %>
 <%- end -%>
-<%- if !Dir[File.join(@metadata[:workdir], 'spec', 'acceptance', '**', '*_spec.rb')].any? && @configs['unit_runs_on'] -%>
+<%- if Dir[File.join(@metadata[:workdir], 'spec', 'acceptance', '**', '*_spec.rb')].none? && @configs['unit_runs_on'] -%>
       unit_runs_on: '<%= @configs['unit_runs_on'] %>'
 <%- end -%>
--- a/moduleroot/.github/workflows/ci.yml.erb
+++ b/moduleroot/.github/workflows/ci.yml.erb
@@ -17,9 +17,15 @@ jobs:
     uses: voxpupuli/gha-puppet/.github/workflows/beaker.yml@v1
     with:
       pidfile_workaround: '<%= @configs['pidfile_workaround'] %>'
+<%- if @configs['unit_runs_on'] -%>
+      unit_runs_on: '<%= @configs['unit_runs_on'] %>'
+<%- end -%>
+<%- if @configs['acceptance_runs_on'] -%>
+      acceptance_runs_on: '<%= @configs['acceptance_runs_on'] %>'
+<%- end -%>
 <%- else -%>
     uses: voxpupuli/gha-puppet/.github/workflows/basic.yml@v1
-<%- if @configs.key?('rubocop') || !@configs['additional_packages'].empty? -%>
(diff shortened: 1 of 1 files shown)


moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -29,3 +35,6 @@ jobs:
 <%- if @configs.key?('rubocop') -%>
       rubocop: <%= @configs['rubocop'] %>
 <%- end -%>
+<%- if !Dir[File.join(@metadata[:workdir], 'spec', 'acceptance', '**', '*_spec.rb')].any? && @configs['unit_runs_on'] -%>
```
ekohl, line 38: You can also use `[].none?` instead of `![].any?`.
```suggestion
<%- if Dir[File.join(@metadata[:workdir], 'spec', 'acceptance', '**', '*_spec.rb')].none? && @configs['unit_runs_on'] -%>
```

moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -17,9 +17,15 @@ jobs:
...
+<%- if @configs['unit_runs_on'] -%>
+      unit_runs_on: '<%= @configs['unit_runs_on'] %>'
+<%- end -%>
+<%- if @configs['acceptance_runs_on'] -%>
+      acceptance_runs_on: '<%= @configs['acceptance_runs_on'] %>'
+<%- end -%>
 <%- else -%>
     uses: voxpupuli/gha-puppet/.github/workflows/basic.yml@v1
-<%- if @configs.key?('rubocop') || !@configs['additional_packages'].empty? -%>
+<%- if @configs.key?('rubocop') || !@configs['additional_packages'].empty? || @configs.key?('unit_runs_on') -%>
```
ekohl, line 28: I wonder at which point we're just going to write it in Ruby and use `to_yaml` on it.

ekohl (approved):

moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -17,9 +17,15 @@ jobs:
     uses: voxpupuli/gha-puppet/.github/workflows/beaker.yml@v1
     with:
       pidfile_workaround: '<%= @configs['pidfile_workaround'] %>'
+<%- if @configs['unit_runs_on'] -%>
+      unit_runs_on: '<%= @configs['unit_runs_on'] %>'
+<%- end -%>
+<%- if @configs['acceptance_runs_on'] -%>
+      acceptance_runs_on: '<%= @configs['acceptance_runs_on'] %>'
```
jhoblitt, line 24: I have been using larger runners in my control repo but I use a nested hash for the config. E.g.:
```
  rake_checks:
    runs-on:
      group: Default Larger Runners
      labels: ubuntu-latest-16-cores
```
I guess that could be done as a heredoc but the indenting would be fiddly.

moduleroot/.github/workflows/ci.yml.erb
bastelfreak: yeah :(

jhoblitt: I've had the same thought. One of the uses with using psych to format is you can't control the indenting of lists and it always uses the non-indented form (which I find harder read).  There's also no support for emitting comments.

moduleroot/.github/workflows/ci.yml.erb
bastelfreak: tested it, works fine

moduleroot/.github/workflows/ci.yml.erb
ekohl: Oh yes, https://github.com/ruby/psych/discussions/594 is still unresolved :(

bastelfreak:
I'm going to merge this since it works. If someone has concrete ideas on how to improve it or want to rewrite it, let me know :)

PR #853, opened 2023-10-06
Add instructions to create a new module

ekohl:


README.md
```diff
@@ -51,6 +51,32 @@ export GITHUB_TOKEN=token
 bundle exec msync update --message "modulesync $(git describe)" --pr --pr-labels modulesync --pr-title "modulesync $(git describe)"
 ```
 
+### Create a new module
+
+It is possible to create a new module using msync.
+First add it to `managed_modules.yml`.
+If it's not in the voxpupuli namespace, be use to include yours.
```
bastelfreak, line 58: `be use`? Is that a correct phrase?

kenyon: I think it's supposed to be

```suggestion
If it's not in the voxpupuli namespace, be sure to include yours.
```

But it could be clearer, like @bschonec said.

What I did was create a branch in my fork that changes the namespace setting, but not sure how maintainable that is: https://github.com/voxpupuli/modulesync_config/compare/master...kenyon:modulesync_config:kenyon

ekohl: You can add `kenyon/puppet-powerwall_dashboard` instead, removing the need to override the namespace in the config.

While writing this I also considered changing modulesync to remove the global setting and simply use full qualified names everywhere.

bin/clean-git-checkouts
```diff
@@ -2,13 +2,13 @@
...
-for module in modules/*/* ; do
-  (cd $module
+for module in modules/voxpupuli/* ; do
+  (cd "$module"
   git status
   git reset --hard
   git clean -dfx
-  git checkout master
-  git pull --prune
+  git checkout "$(basename "$(git symbolic-ref refs/remotes/origin/HEAD)")"
```
bastelfreak, line 10: lol, nice hack 'abusing' basename!

bastelfreak (approved):

kenyon (approved):

README.md
ekohl: > The yml file has pre-loaded modules such as 'puppet-conntrackd'. If I'm trying to create a new module (from scratch?) then is it necessary to remove all of the pre-loaded entries in manage_modules.yml?

No, I used "add" so that implies you don't need to remove other entries.

> Since my module is not in the Puppet forge, what is the correct format to enter? Is it the path to 'git clone' the module?

I've added an example format.

PR #855, opened 2023-10-18
Update to gha-puppet v2

ekohl:
This updates to gha-puppet v2 and raises minimum gem versions so it can support non-AIO Debian 12 testing.


bastelfreak (approved):

ekohl:
Now with GPG signed commit.

PR #864, opened 2024-01-09
Also run CI an the main/master branch

smortex:
When merging a passing PR which was not on top of the last commit of the
main branch, we may break CI.  This make sure a CI run is triggered when
the main branch is updated.  As GitHub kindly display the status of the
last commit on a project page, this allow to quickly see if a module CI
works as intended or not.

The main branch of most of our modules is `master` but a few use the
more inclusive `main` name, so enable ci for both.

### DIFF
--- a/moduleroot/.github/workflows/ci.yml.erb
+++ b/moduleroot/.github/workflows/ci.yml.erb
@@ -4,7 +4,12 @@
 
 name: CI
 
-on: pull_request
+on:
+  pull_request: {}
+  push:
+    branches:
+      - main
+      - master
 
 concurrency:
   group: ${{ github.ref_name }}


moduleroot/.github/workflows/ci.yml.erb
```diff
@@ -4,7 +4,12 @@
 
 name: CI
 
-on: pull_request
+on:
+  pull_request: {}
+  push:
+    branches:
+      - main
+      - master
```
ekohl, line 12: AFAIK we have the remote git branch in modulesync but I'm trying to find out how to use that in a template.

bastelfreak: I think it makes a migration in the future easier if we use both names here? Or is there a chance that a module has a different branch name other than main or master?

ekohl: Turns out we don't: https://github.com/voxpupuli/modulesync/blob/ea247449927d01c8c4fc11c6fd0dbc1e531fbc32/lib/modulesync.rb#L127-L133 does provide `git_base` but https://github.com/voxpupuli/modulesync#configuring-modulesync-defaults does tell the user there's an option for `branch`. That means you could run with `develop` as the default branch.

I can also see a case for making it configurable. In Foreman we also have `*-stable` that we like to run on. So IMHO supporting overriding it in `config_defaults.yaml` would be nice.

smortex:
> Can you test it on [voxpupuli/puppet-example](https://github.com/voxpupuli/puppet-example)?

https://github.com/voxpupuli/puppet-example/pull/39

bastelfreak:
works fine on https://github.com/voxpupuli/puppet-example/actions/runs/7521388462/job/20472166530

smortex:
:wave: I am back from vacations :desert_island: 

I added a config entry as suggested by @ekohl, and checked it was working as expected by running modulesync on the puppet-example module that produce no diff.  Ready for a new review, thanks!

bastelfreak (approved):

PR #866, opened 2024-01-14
CI: Print summary for outdated modules

bastelfreak:
Previously we only printed a diff for every module. Now we also print a hacky table with every outdated module.

### DIFF
--- a/bin/outdated_modules_and_their_version
+++ b/bin/outdated_modules_and_their_version
@@ -1,0 +1,17 @@
+#!/usr/bin/env ruby
+
+require 'yaml'
+require 'erb'
+
+# the current version in modulesync_config
+version = YAML.load(ERB.new(File.read('moduleroot/.msync.yml.erb')).result)['modulesync_config_version']
+
+puts "current version: #{version}"
+puts '---------------------------------------------'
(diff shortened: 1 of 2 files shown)


zilchms (approved):

kenyon (approved):
Table looks good.

bin/outdated_modules_and_their_version
```diff
@@ -0,0 +1,17 @@
...
+version = YAML.load(ERB.new(File.read('moduleroot/.msync.yml.erb')).result)['modulesync_config_version']
+
+puts "current version: #{version}"
+puts '---------------------------------------------'
+puts "| Module\t| modulesync_config version |"
+puts '---------------------------------------------'
+Dir.glob('modules/voxpupuli/puppet-*/.msync.yml').each do|f|
+  version_module = YAML.load_file(f)['modulesync_config_version']
+  mod = f.split('/')[2]
+  puts "| #{mod}\t| #{version_module}\t|" if version != version_module
```
bastelfreak, line 16: I know this is hacky and can be improved. But it provides a readable table and it works without any additional ruby gems.If someone wants to improve it, feel free.

PR #874, opened 2024-02-07
add puppet-sysctl to module sync

crazymind1337:
I have forked thias-sysctl to voxpupuli. This is a module at least I require to be compatible with puppet8. Since the origin hasn't had any development I decided to fork it.


crazymind1337:
Not required anymore.

PR #881, opened 2024-03-13
Fix CONTRIBUTING documentation for modules

stevenpost:
This PR updates the CONTRIBUTING.md file for modules so given commands are actually valid and tests can be run.
The bundler commands given contained deprecated arguments that no longer worked.

### DIFF
--- a/moduleroot/.github/CONTRIBUTING.md.erb
+++ b/moduleroot/.github/CONTRIBUTING.md.erb
@@ -131,19 +131,29 @@ You can install all needed gems for spec tests into the modules directory by
 running:
 
 ```sh
-bundle install --path .vendor/ --without development system_tests release --jobs "$(nproc)"
+bundle config set --local path '.vendor/'
+bundle config set --local without 'development system_tests release'
+bundle install --jobs "$(nproc)"
 ```
 
 If you also want to run acceptance tests:
 
 ```sh
-bundle install --path .vendor/ --with system_tests --without development release --jobs "$(nproc)"
+bundle config set --local path '.vendor/'
+bundle config set --local without 'development release'
+bundle config set --local with 'system_tests'
+bundle install --jobs "$(nproc)"
 ```
 
 Our all in one solution if you don't know if you need to install or update gems:
 
 ```sh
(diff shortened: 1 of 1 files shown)


TheMeier (approved):

moduleroot/.github/CONTRIBUTING.md.erb
```diff
@@ -153,6 +163,11 @@ environment variable:
 BUNDLE_JOBS="$(nproc)"
 ```
 
+Prepare to run tests:
+```sh
```
bastelfreak, line 167: please add an empty line between the sentence and the codeblock (markdown linters demand this).

moduleroot/.github/CONTRIBUTING.md.erb
```diff
@@ -153,6 +163,11 @@ environment variable:
...
+bundle exec rake spec_prep
```
bastelfreak, line 168: Can you explain why you added this? It's not required to run the lint/validate tasks below.

stevenpost: I started out with a new clone of https://github.com/voxpupuli/puppet-mongodb. This command seems to be required when you run `bundle exec rake test`, without running the `spec` rake task first. Although I guess you could say "just run the `spec` task first/instead", I think this is the 'cleaner' approach.

Either way would work for me, I guess.

bastelfreak: can you please try that again? The test rake task is supposed to run the static validations, then spec_prep, then the spec tests.

moduleroot/.github/CONTRIBUTING.md.erb
stevenpost: No longer applicable.

moduleroot/.github/CONTRIBUTING.md.erb
stevenpost: That is indeed not required, my bad. I blame our network, the `spec_prep` task sporadically fails due to network configuration issues (proxy). I dropped the commit.

PR #893, opened 2024-05-05
Cleanup workflow templates; require 'with' key in module's .sync.yml

### LINKED ISSUE #887: Remove tight coupling to gha-puppet in gha workflow templates
When adding new inputs to files in https://github.com/voxpupuli/gha-puppet/tree/v2/.github/workflows we always have to make changes to https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.github/workflows/ci.yml.erb to reflect this.

By adding a 'with' key to .sync.yml and checking `@configs['with']` in ci.yml.erb we could remove this tight copling.

Sample `.sync.yml`
```
.github/workflows/ci.yml:
  with:
    beaker_hosts: 'host1:shard.ma;host2:slave,router.a'
    beaker_facter: 'mongodb_repo_version:MongoDB:4.4,5.0,6.0,7.0'
```

h-haaks:
This is the final stage of removing the tight coupling between reusable gha workflow inputs and the templates in modulesync_config.

Fixes #887

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -2,8 +2,6 @@
 # yamllint disable rule:line-length
 .github/workflows/ci.yml:
   excludes: []
-  pidfile_workaround: false
-  additional_packages: ''
   acceptance_tests: true
   main_branches: ['main', 'master']
 # PDK creates this
--- a/moduleroot/.github/workflows/ci.yml.erb
+++ b/moduleroot/.github/workflows/ci.yml.erb
@@ -21,18 +21,9 @@ jobs:
     name: Puppet
 <%- if @configs['acceptance_tests'] && Dir[File.join(@metadata[:workdir], 'spec', 'acceptance', '**', '*_spec.rb')].any? -%>
     uses: voxpupuli/gha-puppet/.github/workflows/beaker.yml@v2
-<%- if @configs['with'] -%>
-    with:
-<%- @configs['with'].each do |k,v| -%>
-<%- if v.is_a?(String) -%>
-      <%= k %>: '<%= v %>'
-<%- else -%>
-      <%= k %>: <%= v %>
-<%- end -%>
-<%- end -%>
-<%- end -%>
 <%- else -%>
     uses: voxpupuli/gha-puppet/.github/workflows/basic.yml@v2
+<%- end -%>
 <%- if @configs['with'] -%>
     with:
 <%- @configs['with'].each do |k,v| -%>
@@ -40,7 +31,6 @@ jobs:
       <%= k %>: '<%= v %>'
 <%- else -%>
       <%= k %>: <%= v %>
-<%- end -%>
 <%- end -%>
 <%- end -%>
 <%- end -%>
--- a/moduleroot/.github/workflows/ci.yml.erb
+++ b/moduleroot/.github/workflows/ci.yml.erb
@@ -24,7 +24,11 @@ jobs:
 <%- if @configs['with'] -%>
     with:
 <%- @configs['with'].each do |k,v| -%>
+<%- if v.is_a?(String) -%>
       <%= k %>: '<%= v %>'
+<%- else -%>
+      <%= k %>: <%= v %>
+<%- end -%>
 <%- end -%>
 <%- end -%>
 <%- else -%>
@@ -32,7 +36,11 @@ jobs:
 <%- if @configs['with'] -%>
     with:
 <%- @configs['with'].each do |k,v| -%>
+<%- if v.is_a?(String) -%>
       <%= k %>: '<%= v %>'
+<%- else -%>
+      <%= k %>: <%= v %>
+<%- end -%>
 <%- end -%>
 <%- end -%>
 <%- end -%>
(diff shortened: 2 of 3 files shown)


config_defaults.yml
```diff
@@ -2,8 +2,6 @@
 # yamllint disable rule:line-length
 .github/workflows/ci.yml:
   excludes: []
-  pidfile_workaround: false
-  additional_packages: ''
```
h-haaks, line 6: Figured these to be unnecessary as it's the default value defined in the reusable workflows.
The next msync run will remove these inputs from ci.yml.

moduleroot/.github/workflows/release.yml.erb
```diff
@@ -16,10 +16,12 @@ jobs:
     with:
 <%- if @configs['with'] -%>
 <%- @configs['with'].each do |k,v| -%>
-      <%= k %>: <%=  v %>
+      <%= k %>: '<%= v %>'
 <%- end -%>
 <%- end -%>
+<%- if @configs['with'].nil? || !@configs['with'].has_key?('allowed_owner') -%>
       allowed_owner: '<%= @configs[:namespace] %>'
```
h-haaks, line 23: Figure this is whats blocking forks from releasing.
Added a test to prevent adding it double if declared in a modules .sync.yml

h-haaks:
I have used github searches to find all voxpupuli modules using workflow inputs and modified .sync.yml to contain the required 'with' key.
See linked PRs in #887

h-haaks:
Going trough the noop run I see that I have missed a few modules with my searches ...

I'll also have to look into how I write non string values in the templates. i.e. false changes to 'false'

h-haaks:
Seem like I've fixed another issue as well
https://github.com/voxpupuli/modulesync_config/actions/runs/8956855290/job/24599170178?pr=893#step:4:1939
https://github.com/voxpupuli/modulesync_config/actions/runs/8956855290/job/24599170178?pr=893#step:4:6317

h-haaks:
Seems to be ready for review now. I'll fixup my commits before merge.

binford2k:
Hallo @h-haaks! Fill out this form if you'd like to claim your May the Source Be With You sticker! https://forms.office.com/r/Cn55uJmWMH

moduleroot/.github/workflows/ci.yml.erb
ekohl, line 1: I think it can be reduced even further. Sending a diff is tricky, so here's the code:
```erb
jobs:
  puppet:
    name: Puppet
<%- if @configs['acceptance_tests'] && Dir[File.join(@metadata[:workdir], 'spec', 'acceptance', '**', '*_spec.rb')].any? -%>
    uses: voxpupuli/gha-puppet/.github/workflows/beaker.yml@v2
<%- else -%>
    uses: voxpupuli/gha-puppet/.github/workflows/basic.yml@v2
<%- end -%>
<%- if @configs['with'] -%>
    with:
<%- @configs['with'].each do |k,v| -%>
<%- if v.is_a?(String) -%>
      <%= k %>: '<%= v %>'
<%- else -%>
      <%= k %>: <%= v %>
<%- end -%>
<%- end -%>
<%- end -%>
```

You can even use `v.inspect` to avoid the if/else:
```erb
<%- @configs['with'].each do |k,v| -%>
      <%= k %>: <%= v.inspect %>
<%- end -%>
```

moduleroot/.github/workflows/release.yml.erb
```diff
@@ -16,10 +16,16 @@ jobs:
...
 <%- @configs['with'].each do |k,v| -%>
-      <%= k %>: <%=  v %>
+<%- if v.is_a?(String) -%>
+      <%= k %>: '<%= v %>'
+<%- else -%>
+      <%= k %>: <%= v %>
 <%- end -%>
 <%- end -%>
+<%- end -%>
+<%- if @configs['with'].nil? || !@configs['with'].has_key?('allowed_owner') -%>
```
ekohl, line 26: This code is indeed intended to prevent running on forks. Isn't this equivalent?
```suggestion
<%- unless @configs['with']&.has_key?('allowed_owner') -%>
```

moduleroot/.github/workflows/ci.yml.erb
h-haaks: inspect returns a double quoted string. That will generate lots of diffs/changes in the next msync run i guess.
```
irb(main):001:0> t = 'test'
=> "test"
irb(main):002:0> t.inspect
=> "\"test\""
```

moduleroot/.github/workflows/release.yml.erb
h-haaks: Oh, advanced ruby ...
I'll give it a try :)

moduleroot/.github/workflows/ci.yml.erb
ekohl: Oh yes, I may have confused it with Python where `repr(my_string)` is actually using single quotes.

ekohl (approved):
It felt tempting to use `to_yaml`, but I think this is simple enough for now.

PR #895, opened 2024-05-09
Dont manage .rspec_parallel anymore

bastelfreak:
requires https://github.com/puppetlabs/puppetlabs_spec_helper/pull/451 since https://github.com/puppetlabs/puppetlabs_spec_helper/pull/446 the rake task configures the output so we don't need .rspec_parallel anymore.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -16,7 +16,7 @@ Gemfile:
   required:
     ':test':
       - gem: voxpupuli-test
-        version: '~> 7.0'
+        version: '~> 7.2'
       - gem: coveralls
       - gem: simplecov-console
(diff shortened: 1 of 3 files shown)


TheMeier (approved):

ekohl (changes requested):
Can we somehow ensure `puppetlabs_spec_helper` is new enough to include the changes?

bastelfreak:
Yes. I will update voxpupuli-test to depend on the latest PSH as soon as it's released. Then we can depend on the newest voxpupuli-test in modulesync_config.

ekohl (approved):

PR #898, opened 2024-05-21
Use lint rake task for lint pre-commit check

wyardley:
The existing `bundle exec puppet-lint` is prone to erroring with autoload errors that are allowed under our normal structure / lint rules.

```
Runs puppet-lint on modified files only..................[PuppetLint] FAILED
Errors on lines you didn't modify:
/Users/wby/git/puppet-rabbitmq/manifests/init.pp:349:7:ERROR: rabbitmq not in autoload module layout (autoloader_layout)
Warnings on lines you didn't modify:
```

This appears to work from some local testing, and AFAICT still runs only on modified files.


kenyon (commented):
WFM locally too.

kenyon (approved):

PR #899, opened 2024-05-25
.pmtignore: implement module layout specification

bastelfreak:
https://github.com/puppetlabs/puppet-specifications/pull/157 defines which files are allowed in a puppet module release. This commit basically changes the .pmtignore into an allowlist, not a denylist anymore.


moduleroot/.pmtignore.erb
```diff
@@ -1,43 +1,21 @@
...
-/.rubocop.yml
-/.sync.yml
-.*.sw?
-/.yardoc/
-/.yardopts
-/Dockerfile
-/HISTORY.md
-<% if ! @configs['paths'].nil? -%>
-<% @configs['paths'].each do |path| -%>
-<%= path %>
```
bastelfreak, line 41: None of our modules used that

bastelfreak:
I tested this by hand and run `bundle exec module:build` and it worked fine. It would be cooler to implement this also into [puppet-modulebuilder](https://github.com/puppetlabs/puppet-modulebuilder/blob/main/lib/puppet/modulebuilder/builder.rb#L8-L17) but had no luck yet.

bastelfreak:
https://github.com/puppetlabs/puppet-modulebuilder/pull/79 :muscle:

ekohl (commented):
Can we build this list into puppet-modulebuilder? I'm debating how though.

You could make it default, but then it'd still read .gitignore.

You could control it via some option, but we invoke it in 2 ways: release and acceptance tests. We'd need to keep that in sync.

Adding another config file for it is ugly, but enhancing the ignore file with options is ugly too.

rwaffen (approved):

bastelfreak:
this works almost. The problem is that we see empty directories:

```
$ tree puppet-allknowingdns-3.0.2-rc0/
puppet-allknowingdns-3.0.2-rc0/
├── CHANGELOG.md
├── LICENSE
├── manifests
├── metadata.json
├── README.md
├── REFERENCE.md
├── spec
│   ├── acceptance
│   │   └── nodesets
│   ├── classes
│   ├── defines
│   ├── fixtures
│   │   └── manifests
│   ├── functions
│   └── hosts
└── templates
```

bastelfreak:
I raised https://github.com/highb/pathspec-ruby/issues/53

PR #908, opened 2024-06-14
Support fact stringyfication

bastelfreak:
FacterDB/rspec-puppet-facts have an option switch the symbolized first-level keys in a factset to strings (all following/nested keys are always strings). This PR makes this configuration opt-in so we can switch over modules from symbols to strings in their spec files.

### DIFF
--- a/moduleroot/spec/spec_helper.rb.erb
+++ b/moduleroot/spec/spec_helper.rb.erb
@@ -8,9 +8,9 @@
 ENV['COVERAGE'] ||= 'yes' if Dir.exist?(File.expand_path('../lib', __dir__))
 
 require 'voxpupuli/test/spec_helper'
-<%- if @configs['hiera_config'] || @configs['mock_with'] -%>
 
 RSpec.configure do |c|
+  c.facterdb_string_keys = <%= @configs['facterdb_string_keys'] %>
 <%- if @configs['hiera_config'] -%>
   c.hiera_config = <%= @configs['hiera_config'] %>
 <%- end -%>
@@ -18,7 +18,6 @@ RSpec.configure do |c|
   c.mock_with <%= @configs['mock_with'] %>
 <%- end -%>
 end
-<%- end -%>
 <%- if @configs['add_mocked_facts'] -%>
 
 add_mocked_facts!
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -112,6 +112,7 @@ spec/acceptance/nodesets/ec2/windows-2016-base-x64.yml:
   delete: true
 spec/spec_helper.rb:
   add_mocked_facts: true
+  facterdb_string_keys: false
   mock_with: false
 spec/spec_helper_acceptance.rb:
   unmanaged: true


bastelfreak:
Tested at https://github.com/voxpupuli/puppet-allknowingdns/pull/64

ekohl (changes requested):

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -20,6 +21,12 @@ RSpec.configure do |c|
   c.hiera_config = <%= @configs['hiera_config'] %>
 end
 <%- end -%>
+<%- if @configs['facterdb_string_keys'] -%>
```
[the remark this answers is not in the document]
bastelfreak: that's why I added a default to config_defaults.yml. But I can remove the if condition here if you like.

bastelfreak:
verified again at https://github.com/voxpupuli/puppet-allknowingdns/pull/64/commits

moduleroot/spec/spec_helper.rb.erb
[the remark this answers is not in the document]
ekohl: but the default is `false`. If you want it unconditionally, you should check for `.nil?`

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -8,7 +8,7 @@
 ENV['COVERAGE'] ||= 'yes' if Dir.exist?(File.expand_path('../lib', __dir__))
 
 require 'voxpupuli/test/spec_helper'
-<%- if @configs['hiera_config'] || @configs['mock_with'] -%>
+<%- if @configs['hiera_config'] || @configs['mock_with'] || @configs['facterdb_string_keys'] -%>
```
ekohl, line 11: The configure block should be there unconditionally now
```suggestion
```

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -17,6 +17,7 @@ RSpec.configure do |c|
 <%- if @configs['mock_with'] -%>
   c.mock_with <%= @configs['mock_with'] %>
 <%- end -%>
+  c.facterdb_string_keys = <%= @configs['facterdb_string_keys'] %>
 end
 <%- end -%>
```
ekohl, line 22: And then drop this too
```suggestion
```

moduleroot/spec/spec_helper.rb.erb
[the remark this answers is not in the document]
bastelfreak: I'm not sure if I can follow your thoughts. My idea was to always add the `facterdb_string_keys` option with a default of `false` and individual modules can switch to `true`. And if we ever want to, we can set it to `true` in `voxpupuli-test` or `rspec-puppet-facts` and that works with the current code. Did you have something else in mind?

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -17,8 +16,8 @@ RSpec.configure do |c|
 <%- if @configs['mock_with'] -%>
   c.mock_with <%= @configs['mock_with'] %>
 <%- end -%>
+  c.facterdb_string_keys = <%= @configs['facterdb_string_keys'] %>
```
ekohl, line 19: Would it make sense to keep the list sorted alphabetically?

ekohl (approved):

moduleroot/spec/spec_helper.rb.erb
bastelfreak: done

PR #912, opened 2024-06-26
Add archlinux beaker example

traylenator:
Add beaker collection "none" example.

Also update the list of working OSes.


bastelfreak:
@traylenator thanks for the PR! Can you sign you first commit with gpg or an ssh key? This repo requires all commits to be signed.

PR #913, opened 2024-06-28
Also create a GitHub release page upon doing a release

rwaffen:
sadly https://github.com/voxpupuli/gha-create-a-github-release/pull/1 does not work. so simply put the release.yml here, like the labeler.yml.

### DIFF
--- a/moduleroot/.github/release.yml.erb
+++ b/moduleroot/.github/release.yml.erb
@@ -1,0 +1,20 @@
+---
+# https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes
+
+changelog:
+  exclude:
+    labels:
+      - skip-changelog
+
+  categories:
+    - title: Breaking Changes 🛠
+      labels:
+        - backwards-incompatible
+
+    - title: New Features 🎉
+      labels:
+        - enhancement
+
+    - title: Other Changes
+      labels:
+        - "*"
--- a/moduleroot/.github/release.yml.erb
+++ b/moduleroot/.github/release.yml.erb
@@ -21,6 +21,19 @@ changelog:
       labels:
         - enhancement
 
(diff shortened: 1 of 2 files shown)


bastelfreak:
@rwaffen looks good to me, could you run this on our puppet-example module and release it? So we know that it works properly.

rwaffen:
can do

moduleroot/.github/release.yml.erb
```diff
@@ -0,0 +1,20 @@
+---
+# https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes
+
+changelog:
+  exclude:
+    labels:
+      - skip-changelog
```
bastelfreak, line 7: I think `modulesync` should be added here as well? Or everything from https://github.com/voxpupuli/voxpupuli-release/blob/master/lib/voxpupuli/release/rake_tasks.rb#L140 ?

moduleroot/.github/release.yml.erb
```diff
@@ -0,0 +1,20 @@
...
+
+  categories:
```
bastelfreak, line 9: should our `bug` label get its own category?

moduleroot/.github/release.yml.erb
rwaffen: yeah sure, can add all of this.

moduleroot/.github/release.yml.erb
rwaffen: makes sense to have a bug category 🤔

rwaffen:
workflow tested with a release... https://github.com/voxpupuli/puppet-example/releases/tag/v0.4.0

now I will do the requested changes

TheMeier:
nice improvement :)

bastelfreak (approved):

PR #915, opened 2024-07-08
add hint that these files are managed by modulesync

rwaffen:


moduleroot/.puppet-lint.rc.erb
```diff
@@ -1,3 +1,6 @@
+# Managed by modulesync - DO NOT EDIT
```
rwaffen, line 1: here i'am not 100% sure if this works like this

bastelfreak: can you test that please?

rwaffen: yes, will do

rwaffen: seems to work, does not cause an error or change the output

smortex (approved):
Looks reasonable!

PR #920, opened 2024-07-27
Purge devcontainer configs

bastelfreak:
Those files are typically created by pdk and devcontainers cost us money on GitHub.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -156,5 +156,9 @@ Dockerfile:
   delete: true
 .rspec:
   delete: true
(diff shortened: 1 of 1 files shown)


ekohl:
How fo they cost money?

bastelfreak:
I don't know yet how that happend, but we got a 133USD bill. (and we had a spending limit of 50USD, so I realldon't know yet how  that happened, but I'm investigating)

The code spaces allocate disk space and that's billed by GitHub.

ekohl (approved):
That is weird. Let's get rid of them then.

PR #933, opened 2024-09-16
managed_modules.yml: sort modules alphabetically

bastelfreak:


jhoblitt (approved):
Maybe we should add some rspec to check that the list is sorted?

smortex (approved):

kenyon:
Note that order depends on how the sort is implemented. For example, Visual Studio Code's built-in sort function works differently than the coreutils `sort` command: https://github.com/microsoft/vscode/issues/48123

So it would be good to have a standard way of doing it, like with a rake task, and a verification method, like @jhoblitt says.

kenyon (approved):

jhoblitt:
I have a close enough copy of coreutils sorting behavior to use it in a spec file. https://github.com/lsst-it/lsst-control/blob/production/spec/unit/puppetfile_spec.rb

PR #934, opened 2024-10-01
puppet-catalog-diff: the module was renamed

bastelfreak:
The module got renamed from puppet-catalog-diff to puppet-catalog_diff


kenyon (approved):
Can confirm: https://github.com/voxpupuli/puppet-catalog_diff/blob/0ab900021c82beda3dcb6c44cfcfb138ccb52866/metadata.json#L2

PR #950, opened 2025-03-19
Add GitHub Action to run modulesync

bastelfreak:
This was tested in
https://github.com/voxpupuli/puppet-lint_modulesync_configs. The new action will run `msync update` for each new tag, or on demand. It will also create PRs in all managed modules. existing `modulesync` branches will be updated.


.github/workflows/update.yml
```diff
@@ -0,0 +1,60 @@
...
+    if: github.repository_owner == voxpupuli
+    steps:
+      - name: checkout code
+        # the action has a ssh-key option.
+        # when provided, it configures the ssh/git config to authenticate automatically with that key to the source repo
+        # the authentication won't be used for other repos (which we need for msync), so we cannot use the option
+        # that's the reason for the 'Add SSH key' step
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
ekohl, line 31: Do we need the entire history to be fetched? I wouldn't think modulesync cares about that

.github/workflows/update.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+            mkdir -p ~/.ssh
+            echo "${{ secrets.PCCI_SSH_PRIVATE_KEY }}" > ~/.ssh/github_actions
+            chmod 600 ~/.ssh/github_actions
+            ssh-agent -a $SSH_AUTH_SOCK > /dev/null
+            ssh-add ~/.ssh/github_actions
+      - name: set git user & email & commit signing
+        run: |
+          git config --global user.email '12855858+pccibot@users.noreply.github.com'
+          git config --global user.name 'pccibot'
```
ekohl, line 47: You set `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL` so isn't this redundant? If so:
```suggestion
```

Or is it needed for signing? Can we then reduce duplication
```suggestion
          git config --global user.email "$GIT_AUTHOR_EMAIL"
          git config --global user.name "$GIT_AUTHOR_NAME"
```

.github/workflows/update.yml
```diff
@@ -0,0 +1,60 @@
+---
+name: Update all puppet modules
```
ekohl, line 2:
```suggestion
name: Update all Puppet modules
```

sebastianrakel (approved):

ekohl (commented):
Nice idea. I toyed with this idea before.

If there's already a PR with an existing branch it will only add a commit, not drop any manual work that's been added, correct?

bastelfreak:
> If there's already a PR with an existing branch it will only add a commit, not drop any manual work that's been added, correct?

yes correct

.github/workflows/update.yml
bastelfreak: I didn't understans why we need the env vars and the updated git config, but there were issues when either was missing

.github/workflows/update.yml
ekohl: https://github.com/actions/checkout mentions there's `fetch-tags` so I suspect this works as well:
```suggestion
        with:
          fetch-tags: true
```

bastelfreak: I need to test this, I think that only works when the action runs for a pushed tag, not for the on demand runs.

.github/workflows/update.yml
```diff
@@ -0,0 +1,60 @@
+---
+name: Update all Puppet modules
+
+on:
```
ekohl, line 4: Something that occurred to me when reviewing https://github.com/voxpupuli/modulesync_config/pull/951: we can probably also make it trigger on file changes to `managed_modules.yml`. It may need its own workflow that checks out the latest tag instead of the default branch, but can help reduce some manual effort in onboarding new modules.

PR #957, opened 2025-04-30
Automatically label modulesync PRs accordingly

smortex:
When using modulesync, we regularly end-up creating many pull-requests
to merge a `modulesync` branch into main / master.

We have a `modulesync` label that allows such PRs to be excluded from
the ChangeLog… but forgetting to label them when we create them in batch
mean we have to label them one by one by hand which is a PITA.

Adjust the labeller configuration so that PRs against a branch named
`modulesync` is automatically tagged `modulesync`.


bastelfreak (approved):

bastelfreak:
We also have https://github.com/voxpupuli/modulesync_config/blob/master/.github/workflows/update.yml, which runs msync on all modules and attaches the correct label.

PR #961, opened 2025-05-28
voxpupuli-test: Update 9.x -> 10.x

bastelfreak:


config_defaults.yml
```diff
@@ -16,7 +16,7 @@ Gemfile:
   required:
     ':test':
       - gem: voxpupuli-test
-        version: '~> 9.0'
+        version: '~> 10.0'
       - gem: coveralls
       - gem: simplecov-console
```
ekohl, line 21: We dropped coveralls integration so should we also drop this?
```suggestion
```

PR #965, opened 2025-05-28
Release CI: Checkout code before creating release

bastelfreak:


rwaffen:
can you also please add a release config like in https://github.com/voxpupuli/container-voxbox/blob/main/.github/release.yml here?

rwaffen (approved):

PR #966, opened 2025-05-29
CI jobs: Add explicit token permissions

bastelfreak:
Tested in https://github.com/voxpupuli/puppet-example/pull/117

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -7,6 +7,9 @@ on:
     tags:
       - '*'
 
+permissions:
+  contents: read
+
 env:
   GIT_AUTHOR_NAME: pccibot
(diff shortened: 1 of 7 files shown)


.github/workflows/update.yml
```diff
@@ -7,6 +7,9 @@ on:
     tags:
       - '*'
 
+permissions:
+  contents: read
```
jpartlow, line 11: Does this need permission to write a pull request?

bastelfreak: short: no.

long: to prevent recursion, a github token can raise a PR, but it won't trigger any CI runs on that PR. that's why we have a very annoying setup with a github bot account, https://github.com/pccibot, which has a token and we use that token in line 57. And it only took a week to figure that out.

jpartlow: whacky

PR #967, opened 2025-06-11
Update voxpupuli-release/voxpupuli-test to use openvox gems

bastelfreak:


ekohl (changes requested):
We need to update this too:
https://github.com/voxpupuli/modulesync_config/blob/57ff47b96c91a8bd9461c2fc1976ee3f3cf138f2/moduleroot/Gemfile.erb#L38-L39

bastelfreak:
should we stick with `PUPPET_GEM_VERSION` or also update it? That would mean we've to do a new release for gha-puppet. Or do something like `ENV['OPENVOX_VERSION'] || ENV.fetch('PUPPET_VERSION', '>= 7.24')`

d1nuc0m:
Seems ok to me, as [ openvox gem](https://rubygems.org/gems/openvox) already depends on facter it shouldn't be a problem to remove it

PR #969, opened 2025-06-15
feat: Add cleanup script for fixtures

TheMeier:
Fixes: #48

I have separated out the git clone-or-update in a separate file, since this is already a duplication from add_openvox_to_metadata_json

This generates quite some noise, because it strips all comments and all unnecessary quotes in `.fixtures.yml`. One could argue that is a positive since everything is homogeneous afterwards. I guess handling comments would be very because of there positional nature. Also there is not many files with comments and they mostly are useless anyway.

Most often its
```
# This file can be used to install module dependencies for unit testing
# See https://github.com/puppetlabs/puppetlabs_spec_helper#using-fixtures for details
```
We could consider a static common header if wanted

I personally also think it is not worth it to handle the optional quotes, but I guess that can be achieved. But not without some effort


lib/hash.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+def deeply_sort_hash(object)
+  return object unless object.is_a?(Hash)
+
+  hash = {}
+  object.each { |k, v| hash[k] = deeply_sort_hash(v) }
+  sorted = hash.sort { |a, b| a[0].to_s <=> b[0].to_s }
+  hash.class[sorted]
```
ekohl, line 9: why `hash.class[sorted]`? Is this to instantiate a new class?

lib/hash.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+def deeply_sort_hash(object)
+  return object unless object.is_a?(Hash)
+
+  hash = {}
+  object.each { |k, v| hash[k] = deeply_sort_hash(v) }
+  sorted = hash.sort { |a, b| a[0].to_s <=> b[0].to_s }
```
ekohl, line 8: I think this should work too
```suggestion
  hash = object.transform_values { |v| deeply_sorted_hash(v) }.sort_by(&:to_s)
```

ekohl:
There is `msync exec` too. Could we reuse that to avoid rewriting a lot of the msync logic?

TheMeier:
> There is `msync exec` too. Could we reuse that to avoid rewriting a lot of the msync logic?

well actually I was wondering if there was already some "plugin" style feature for modulesync to run snippets/includes on all repos. But I just looked at https://github.com/voxpupuli/community-triage/issues/48.

I agree that this should rather be implement as a script that is designed for `msync exec` 
@bastelfreak ?

PR #980, opened 2025-08-11
voxpupuli-acceptance: Require 4.x

bastelfreak:
This pulls in a new beaker major release, which requires Ruby 3.2.


bastelfreak:
tested in https://github.com/voxpupuli/puppet-example/pull/130

PR #989, opened 2025-08-19
Add a utility script to bump supported OS versions

smortex:
Add support for a newer version of an Operating System if the previous
version is supported.  Update metadata.json, create a feature branch,
commit, and open a PR automatically.

If the previous version was not supported but older version are, prompt
the user interactively for what to do.

If the OS is not supported at all, do nothing.

Typical usage (when Debian 13 is released):

```
bundle exec msync execute -B -- git pull
bundle exec msync execute -B -- $PWD/bin/add-support-for-new-os -o Debian -v 13
```

### DIFF
--- a/bin/add-support-for-new-os
+++ b/bin/add-support-for-new-os
@@ -1,0 +1,151 @@
+#!/usr/bin/env ruby
+#
+# Add support for a newer version of an Operating System if the previous
+# version is supported.  Update metadata.json, create a feature branch, commit,
+# and open a PR automatically.
+# If the previous version was not supported but older version are, prompt the
+# user interactively for what to do.
+# If the OS is not supported at all, do nothing.
+#
+# Typical usage (when Debian 13 is released):
+#
+#     bundle exec msync execute -B -- git pull
+#     bundle exec msync execute -B -- $PWD/bin/add-support-for-new-os -o Debian -v 13
+
+require 'json'
+require 'optparse'
+
+require 'tty-prompt'
+
+class Options
+  attr_accessor :os, :version
+  attr_writer :noop, :interactive
+
+  def initialize
+    @noop = false
+  end
+
+  def branch_name
+    @branch_name ||= "#{os.downcase}-#{version}"
+  end
+
+  def noop?
+    @noop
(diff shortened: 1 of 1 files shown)


bin/add-support-for-new-os
```diff
@@ -0,0 +1,151 @@
...
+  r, w = IO.pipe
+  w.puts(pp_metadata)
+  w.close
+  system "diff", "-u", "metadata.json", "-", in: r
+else
+  system "git", "checkout", "-b", options.branch_name
+  File.write('metadata.json', "#{pp_metadata}\n")
+  system "git", "add", "metadata.json"
+  system "git", "commit", "-m", "Add support for #{options.os} #{options.version}"
+  system "gh", "pr", "create", "--fill", "--draft", "--label", "enhancement"
```
ekohl, line 150: There is `msync exec` for that. If it doesn't work, then we should enhance it so it does work.

bin/add-support-for-new-os
```diff
@@ -0,0 +1,151 @@
...
+    exit 0
+  end
+  puts
+end
+
+puts(pastel.green("Adding support for #{options.os} #{options.version}..."))
+
+os_index = metadata['operatingsystem_support'].map { |os| os['operatingsystem'] }.index(options.os)
+
+metadata['operatingsystem_support'][os_index]['operatingsystemrelease'] << options.version
```
ekohl, line 136: `operatingsystemrelease` can be `nil` and it doesn't deal with that. This is also fragile because it doesn't handle a missing OS (typo).

If anything you should probably do something like:

```suggestion
os = metadata['operatingsystem_support'].find { |os| os['operatingsystem'] == options.os }
unless os
  exit 1
end
if os['operatingsystemrelease']
  os['operatingsystemrelease'] << options.version
else
  os['operatingsystemrelease'] = [options.version]
end
```

ekohl (changes requested):
I don't think this belongs in our modulesync config. I wrote https://github.com/voxpupuli/puppet_metadata/pull/188 which already knows about which OS releases could/should be added.

PR #990, opened 2025-08-26
Add the puppet-cups module

smortex:


kenyon (approved):
https://github.com/voxpupuli/puppet-cups/issues/538

PR #1003, opened 2025-10-17
Add allowed_owner setting to avoid changes on fork

Joris29:
This allows people to use modulesync_config to run msync update on forked projects without changing GitHub related things.

Example of current behaviour:
image

Example of new behaviour:
image


Joris29:
Yeah sure

